PR #5, opened 2019-11-27
Add Jenkins job info along with Jenkins script for CD

sidpremkumar:
TODO: Add jenkins information

So far this PR:
- Adds Script to call Sync2Jira and then compares tickets with pre-computed values


sidpremkumar:
I don't think I want to use Jenkins anymore. I was thinking more of a simple app that we could deploy in the Sync2Jira namespace that would utilize this: 
https://developer.github.com/v3/guides/building-a-ci-server/

sidpremkumar:
Closing this PR in favor of #27

PR #59, opened 2020-01-06
Add comment when fixVersion is unable to be updated

sidpremkumar:
Indicate on the downstream issue if the FixVersion was unable to be updated.


sidpremkumar:
Relates to JIRA: DEVOPSA-7263

PR #66, opened 2020-01-27
Add mailing list support to send failure emails

sidpremkumar:
Send failure emails to the mailing list instead of admins.


sidpremkumar:
Relates to JIRA: DEVOPSA-7333

PR #83, opened 2020-03-03
Continuous deployment

sidpremkumar:
Merging the CD branch into master so we can keep them the same. This PR is all the changes I made while deploying. 

Left commits because it indicates better what I was working on then a single squashed commit but open to changing that.

### DIFF
--- a/continuous-deployment/install_rhmsg.sh
+++ b/continuous-deployment/install_rhmsg.sh
@@ -1,19 +1,29 @@
 #!/usr/bin/bash
 
-if [ -n "$rcm_tools_repos" ]; then
-    repo_file=/etc/yum.repos.d/rcm-tools-fedora.repo
-    curl -L -o $repo_file $rcm_tools_repos
-    # Since we don't trust any internal CAs at this point, we must connect over
-    # http
-    sed -i 's/https:/http:/g' $repo_file
+main() {
+  if [ -n "$RCM_TOOLS_REPO" ]; then
+      repo_file=/usr/local/src/sync2jira/continuous-deployment/rcm-tools-fedora.repo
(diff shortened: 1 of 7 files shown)


sidpremkumar:
Relates to JIRA: DEVOPSA-7103

continuous-deployment/install_rhmsg.sh
```diff
@@ -1,19 +1,29 @@
...
-    # Since we don't trust any internal CAs at this point, we must connect over
-    # http
-    sed -i 's/https:/http:/g' $repo_file
+main() {
+  if [ -n "$RCM_TOOLS_REPO" ]; then
+      repo_file=/usr/local/src/sync2jira/continuous-deployment/rcm-tools-fedora.repo
+      curl -L -o $repo_file $RCM_TOOLS_REPO
+      # Since we don't trust any internal CAs at this point, we must connect over
+      # http
+      sed -i 's/https:/http:/g' $repo_file
```
ralphbean: What prevents us from trusting the internal CA before this point?

sidpremkumar: We add the internal CA here: https://github.com/sidpremkumar/Sync2Jira/pull/83/files#diff-9d8e13c2dac78d2e51e7ad1eebff3f95R52

So prior to this, before the container starts, we don't have the CA_URL needed to get the certs. 

Used this as a guide: https://github.com/fedora-modularity/message-tagging-service/blob/master/docker/install-dependencies.sh

ralphbean: Ack.  We trust the CA at runtime, but here we are working at build time.

PR #91, opened 2020-03-31
Catch connection timeout error

sidpremkumar:
Sync2Jira keeps failing due to a connection timeout with github. Apparently it does not actaully try to re-try it, it just tried it once and fails. When we get a connection timeout, we should wait, and re-sync whatever we were syncing. 

Also, fix a small assignee error

https://github.com/psf/requests/issues/1198

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -1034,7 +1034,7 @@ def _update_assignee(client, existing, issue, updates):
                 confluence_data = {'Assignee': 1}
                 confluence_client.update_stat_page(confluence_data)
         else:
-            if existing.fields.assignee and existing.fields.assignee.displayName != issue.assignee[0]['fullname']:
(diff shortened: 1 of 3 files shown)


sync2jira/downstream_issue.py
```diff
@@ -1034,7 +1034,7 @@ def _update_assignee(client, existing, issue, updates):
                 confluence_data = {'Assignee': 1}
                 confluence_client.update_stat_page(confluence_data)
         else:
-            if existing.fields.assignee and existing.fields.assignee.displayName != issue.assignee[0]['fullname']:
+            if existing.fields.assignee and not issue.assignee:
```
sidpremkumar: No one uses assignee so that why it wasn't caught, but I noted this did not make sense so I fixed it in this PR as well.

ralphbean: Nice.  :)

PR #101, opened 2020-04-05
Remove failures from happening inside confluence_client.py

sidpremkumar:
Relates to JIRA: DEVOPSA-7698
Right now, oddly, we get issues posting or getting from confluence (few and far between). The confluence update was indented for statistical and informational purposes and should not be able to take down and restart the sync2jira service. 

I propose removing and `resp.raise_for_status()` code from `confluence_client.py`. If it can't sync and update confluence log the problem and continue, do not restart. It's okay if we don't update all the stats, it is not critical.

### DIFF
--- a/sync2jira/confluence_client.py
+++ b/sync2jira/confluence_client.py
@@ -12,7 +12,7 @@ import jinja2
 import datetime
 
 # Global Variables
-log = logging.getLogger(__name__)
+log = logging.getLogger('sync2jira')
 
 
 class ConfluenceClient:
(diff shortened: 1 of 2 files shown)


sync2jira/confluence_client.py
```diff
@@ -78,93 +78,96 @@ def update_stat_page(self, confluence_data):
         Updates the statistic page with more data
         :param dict confluence_data: Variable amount of new data
         """
-        # Get the HTML to update
```
sidpremkumar, line 81: I just placed all this deleted code inside a try/catch

sidpremkumar:
@ralphbean bump

sync2jira/confluence_client.py
```diff
@@ -78,93 +78,96 @@ def update_stat_page(self, confluence_data):
...
+                searchpath='usr/local/src/sync2jira/sync2jira/')
+            templateEnv = jinja2.Environment(loader=templateLoader)
+            template = templateEnv.get_template('confluence_stat.jinja')
+            html_text = template.render(confluence_data=confluence_data_update)
+
+            # Finally update our page
+            if html_text.replace(" ", "") != page_html.replace(" ", ""):
+                self.update_page(self.page_id, html_text)
+        except Exception as e:
+            log.error(f"Something went wrong updating confluence!\nException: {e}")
```
ralphbean, line 170: Use `log.exception(...)`.  The nice thing about it is that it logs the whole traceback, which is useful for debugging.

Just `log.exception("Something went wrong updating confluence!")` should be adequate.

PR #105, opened 2020-05-15
Add support to apply labels when upstream issue is closed

csomh:
Make it possible to configure a list of labels to be applied on Jira
cards, when the corresponding upstream issue is closed.

Example:

```python
{
    'issue_updates': [
        {'on_close': {'apply_labels': ['closed-upstream']}
    ]
}
```

This can come handy, when one doesn't want to automatically close
downstream issues when upstream issues are closed, but still want's to
have an easy way to filter for such issues in Jira.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -718,7 +718,7 @@ def _update_jira_issue(existing, issue, client):
     log.info("Updating information for upstream issue: %s" % issue.title)
 
     # Get a list of what the user wants to update for the upstream issue
-    updates = issue.downstream.get('issue_updates', {})
+    updates = issue.downstream.get('issue_updates', [])
 
     # Update relevant data if needed
     # If the user has specified nothing
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -1060,6 +1060,27 @@ def _update_assignee(client, existing, issue, updates):
                     confluence_client.update_stat_page(confluence_data)
 
 
+def _update_jira_labels(issue, labels):
+    """Update a Jira issue with 'labels'
+
+    Do this only if the current labels would change.
+
+    :param jira.resource.Issue issue: Jira issue to be updated
+    :param list<strings> labels: Lables to be applied on the issue
+    :returns: None
+    """
+    _labels = sorted(labels)
+    if _labels == sorted(issue.fields.labels):
+        return
+
+    data = {'labels': _labels}
+    issue.update(data)
+    log.info('Updated %s tag(s)' % len(_labels))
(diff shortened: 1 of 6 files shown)


sidpremkumar:
Has this been tested locally?

csomh:
> Has this been tested locally?

I don't have the environment to do so, so it wasn't.

sync2jira/downstream_issue.py
```diff
@@ -767,6 +767,16 @@ def _update_jira_issue(existing, issue, client):
         log.info("Looking for new transition(s)")
         _update_transition(client, existing, issue)
 
+    on_close_updates = None
```
sidpremkumar, line 770: Could you add a comment on what we are syncing `# Only synchronize close labels for listings tat op-in?

Also, can we move lines 770-774 to inside the `_update_on_close` function? That way we can keep `_update_jira_issue` really clean?

sidpremkumar:
> > Has this been tested locally?
> 
> I don't have the environment to do so, so it wasn't.

Okay, when we are ready to merge. I will test locally prior.

sync2jira/downstream_issue.py
csomh: > Could you add a comment on what we are syncing `# Only synchronize close labels for listings that op-in?`
> 

Sure.

> Also, can we move lines 770-774 to inside the `_update_on_close` function? That way we can keep `_update_jira_issue` really clean?

Doing this would mean that the `issue_updates` list would be needed to be parsed 2 times: once in `if any(..)` and then once again in `_update_on_close()`, in order to extract the `on_close` dictionary from the list. Is doing this work twice okay?

csomh: Or `_update_on_close()` could be called without a condition. But then it would not be visible from `_update_jira_issue` that this action as optional.

sidpremkumar: Call `_update_on_close()` without a conditional, log something like `Attempting to sync close labels` and then inside `_update_on_close ()` log something like `Synced XYZ labels` if we were able to sync anything. 

This has been the workflow we've used for other fields.

sidpremkumar: If you think it makes more sense, you can also check the conditional first. I will leave it up to you, but try to take as much code off our `_update_jira_issue`

csomh: Updated to have an unconditional `_update_on_close()` call.

sidpremkumar: Looks good, I'll test & merge if it passes later today.

sidpremkumar:
It looks like its working. Merging: 
image

PR #110, opened 2020-05-26
Update CI to include sync-page

sidpremkumar:


sidpremkumar:
@gabdzo  Just wanted to show you this, this is how we add CD items, FYI: we build this service (image) on openshift, not through quay as it requires private information.

PR #112, opened 2020-05-26
Add issue links to pr

sidpremkumar:
~~Testing on Stage prior to master~~
Tested on Stage and everything went as expected. 
From the original issue:

> It would be cool when we link a PR we add the PR URL to issue_links on JIRA
> Taking this Issue. also we need to check against issue_links, instead of just looking for a matching comment. I noticed that sometimes we could double comment when a user would remove the WIP title.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -434,13 +434,13 @@ def _get_existing_jira_issue_legacy(client, issue, config):
         return None
 
 
-def _attach_link(client, downstream, remote_link):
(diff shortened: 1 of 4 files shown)


sidpremkumar:
@gabdzo PTAL as well , thanks

sidpremkumar:
Relates to JIRA: PVDRASP-554

ralphbean:
I'll leave this one to @gabdzo.  :)

gabdzo:
+1

PR #113, opened 2020-11-11
Convert Pagure description from Markdown to Jira format

hluk:
Fixes also following exception:

    TypeError: not all arguments converted during string formatting

### DIFF
--- a/sync2jira/intermediary.py
+++ b/sync2jira/intermediary.py
@@ -19,6 +19,13 @@
 from datetime import datetime
 import re
 
+import pypandoc
+
+
+def _md_to_jira(markdown):
+    """Convert Markdown to Jira"""
+    return pypandoc.convert_text(markdown, 'jira', format='markdown').rstrip('\n')
+
 
 class Issue(object):
     """Issue Intermediary object"""
@@ -75,7 +82,7 @@ class Issue(object):
             comment['date_created'] = datetime.fromtimestamp(float(comment['date_created']))
             comments.append({
                 'author': comment['user']['name'],
-                'body': comment['comment'],
+                'body': _md_to_jira(comment['comment']),
                 'name': comment['user']['name'],
                 'id': comment['id'],
                 'date_created': comment['date_created'],
@@ -99,7 +106,7 @@ class Issue(object):
             tags=issue['tags'],
             fixVersion=[issue['milestone']],
             priority=issue['priority'],
-            content=issue['content'],
+            content=_md_to_jira(issue['content']),
             reporter=issue['user'],
             assignee=issue['assignee'],
             status=issue['status'],
@@ -221,7 +228,7 @@ class PR(object):
                 float(comment['date_created']))
             comments.append({
                 'author': comment['user']['name'],
-                'body': comment['comment'],
+                'body': _md_to_jira(comment['comment']),
                 'name': comment['user']['name'],
                 'id': comment['id'],
                 'date_created': comment['date_created'],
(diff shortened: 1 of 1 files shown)


hluk:
It would be also safer to use `{panel:title=Upstream Description}...{panel}` instead of `{quote}...{quote}` because quotes in the description or comment would break the formatting.

sync2jira/intermediary.py
```diff
@@ -19,6 +19,13 @@
 from datetime import datetime
 import re
 
+import pypandoc
```
sidpremkumar, line 22: Make sure to add this to the [DockerFile](https://github.com/release-engineering/Sync2Jira/blob/master/Dockerfile) if it needs to be pip installed.

hluk: It's already installed using `dnf`. The package is used to convert github comments to plain text (based on the `github_markdown` option).

ralphbean:
Can you add the following to the travis file?

```
before_install:
- sudo apt-get install -qq pandoc
```

hluk:
> Can you add the following to the travis file?
> 
> ```
> before_install:
> - sudo apt-get install -qq pandoc
> ```

Unfortunately, pandoc in the latest ubuntu (20.04) available on Travis CI is too old (version 2.5) and doesn't contain jira output format.

Using Github Actions would work, but it's bigger change. This would also speed up the builds I think - Travis CI is now very slow.

Any other ideas how to use newer pandoc for tests?

hluk:
I'm trying to get the GitHub Actions working for this: #114

ralphbean:
Weird - any luck with github actions?

hluk:
Unfortunately, I haven't got much time to look into this. Perhaps the patch needs to be merged to master to start working.

@sidpremkumar Any idea if something is overriding the CI in this github project? I'm not sure where "Sync2Jira Build" test is coming from.

Replacing the Travis CI is needed because it's [going away at the end of the year](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom):

> ... we are announcing that travis-ci.org will be officially closed down completely on December 31st, 2020 ...

sidpremkumar:
Oh wow, I didn't know Travis was closing. 

Sync2Jira build is coming from settings -> Integrations. It's set up as a webhook that pushes to quay. Then there is an OpenShift instance running that listens to that quay webhook, performs some integration tests, and then pushes back up. 

I'm not sure if I have access to it anymore, but if I do feel free to ping me and we can get it sorted. 

Let me know if there is anything else I can help with. Would love to contribute any way I can.

sidpremkumar:
I believe the reason the sync2jira build is faling is because the Jira tickets are not synced correctly. There should be a config file in OpenShift that maps the upstream issues to downstream JIRA tickets. But every so often the JIRA tickets get updated and the numbering gets thrown out of wack. See [this](https://github.com/release-engineering/Sync2Jira/pull/113/checks?check_run_id=1425095665).

If its too much trouble and not providing much utility (which i feel it might), might be best to just remove sync2jira build entierly. I just put it there becuase when I was working on Sync2Jira I found is useful to have some integration tests before merging to master.

hluk:
Updated. Now the tests pass.

@sidpremkumar Looks like currently the integration tests (failing Sync2Jira Build) only compare values in existing Jira issues. This doesn't seem to be useful but maybe I'm missing something.

sidpremkumar:
@hulk yes, I would probably remove them or spend some time and fix them. I worked on them when I had some free time during my internship. More of a test project vs. something super useful. If you choose to fix them I would be more than glad to help work on that.

hluk:
Is this OK to merge?

Issue #123, opened 2021-09-08
Error hit when syncing GH repo which contains comment with more than 100000 characters


tplevko:
I hit this issue with this GH issue's comment - https://github.com/apache/camel-quarkus/issues/2957#issue-955186387, with more than 180K characters

it's first comment, so it was supposed to be used as a description in the Jira issue - this doesn't work, as the Jira API description limit is 1000000 characters.

Could this be limited on the sync2jira side?

tplevko:
Edit: seems like the limit is 65,534 characters:

```response text = {"errorMessages":[],"errors":{"description":"The entered text is too long. It exceeds the allowed limit of 65,534 characters."}}```

**Comment by tplevko** [comment]
@sidpremkumar could you please have a look?

**Comment by tplevko** [comment]
@sidpremkumar sounds great! I'll ask someone if this can be tested - I don't have access to the deployment myself.

**Comment by sidpremkumar** [comment]
I've created this could you take a look: https://github.com/release-engineering/Sync2Jira/pull/124

**Comment by sidpremkumar** [comment]
Believe this issue can be closed now?

**Comment by qgong** [comment]
Close it.

**Comment by tplevko** [comment]
Seems like the fix didn't prevent the issue from happening - I can still see the error from the first comment in the pod logs. Do you maybe have an idea why the patch didn't solve the issue?

**Comment by sidpremkumar** [comment]
@tplevko , just to confirm how do we know it's coming from a comment?

**Comment by tplevko** [comment]
@sidpremkumar placing the full error log: 

```
[2021-10-05 23:35:19,468] INFO: Creating {'project': 'CQ', 'labels': ['upstream'], 'issue_updates': ['title', 'description', 'comments'], 'sync': ['issue'], 'jira_instance': 'omegaprime-jira'} issue for <Issue https://github.com/apache/camel-quarkus/issues/2957 >
[2021-10-05 23:35:19,468] INFO: Creating issue.
[2021-10-05 23:35:19,881] ERROR:    Failed on <Issue https://github.com/apache/camel-quarkus/issues/2957 >
[2021-10-05 23:35:22,833] ERROR: Exception on /handle-event [POST]
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.9/site-packages/flask/app.py", line 1952, in full_dispatch_request
[...]

PR #124, opened 2021-09-22
Add logic to trim long comments

sidpremkumar:
Adds logic to trim comments to ensure that we do not hit JIRA length limits. 

Relates to #123


qgong:
LGTM

ralphbean:
Looks good. Thanks @sidpremkumar, @qgong!

PR #127, opened 2023-04-14
Docs: update documentation

zdtsw:
- add "initialize" into config map as example
- correct to use "True" than "1" to match test and README
- add missing "," in config
- format with indentation


ralphbean:
This is great; thanks Wen!

ralphbean (approved):

PR #134, opened 2023-09-29
Use jira format and gfm when converting desc

lcarva:
While syncing an issue to Jira, if the `github_markdown` option is listed in `issue_updates`, and the upstream issue is from GitHub, then use the proper GitHub-Flavored Markdown parser (`gfm`) and covert it to the `jira` format instead of `plain` text.

This improves how things like links and code blocks are represented in the created Jira issue.

### DIFF
--- a/docs/source/config-file.rst
+++ b/docs/source/config-file.rst
@@ -141,7 +141,7 @@ The config file is made up of multiple parts
     * :code:`{'on_close': {'apply_lables': ['label', ...]}}`
         * When the upstream issue is closed, apply additional labels on the corresponding Jira ticket.
     * :code:`github_markdown`
-        * If description syncing is turned on, this flag will convert Github markdown to plaintext. This uses the pypandoc module.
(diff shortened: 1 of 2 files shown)


lcarva:
I was able to verify that the pypandoc version in the image does have those formats:

```bash
🐚 podman run --rm -it --entrypoint python3 quay.io/redhat-aqe/sync2jira:latest
Python 3.11.4 (main, Jun  7 2023, 00:00:00) [GCC 13.1.1 20230511 (Red Hat 13.1.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pypandoc
>>> pypandoc.convert_text('[foo](https://foo.com) `bar`', 'jira', format='gfm')
'[foo|https://foo.com] {{bar}}\n'
>>> 

```

lcarva:
@ralphbean, anything I can do to help this get merged?

ralphbean:
Oop! Bugging me helps. Let's merge it.

PR #175, opened 2024-07-23
implement a debug logging config option

Zyzyx:
Add a _required_ `debug` option to the configuration file. It will turn on or off debugging output.


ralphbean:
Let's get this one merged after the others make it to prod, just so we can be sure we're changing only one thing at a time.

PR #193, opened 2024-07-26
Use raw strings for flake8

ralphbean:
flake8 was choking on these lines.


ralphbean:
Before merging this, we should make sure we have a test that covers that function. Is this breaking anything?

ralphbean:
OK, I confirmed that zero tests cover this function - which means this is risky.

I'll write some tests that pass on the old code first, and then confirm they work on the new code.

ralphbean:
OK, I feel better about it now.

PR #195, opened 2024-07-26
Drop unused openshift configs

ralphbean:
These aren't used anywhere now.


ralphbean:
Ah, I was wrong. one thing is used in the dockerfile.

PR #203, opened 2024-09-16
Basic implementation of storypoints sync

### LINKED ISSUE #165: Sync Story Points
As title says, enable sync'ing story points from GitHub to story points in Jira.

baijum:
Fixes #165

An example configuration:

```
open-connectors/open-connectors': {'project': 'TEST', 'component': 'kernel',
   'updates': ['pullrequest', 'issue'],
  'issue_updates': ['comments', 'github_project_fields'],
  'github_project_fields': {'storypoints': ('Estimate', 'customfield_12316544')},
  'sync': ['pullrequest', 'issue']},
```

And default config:
```
 'default_github_project_fields': {'storypoints': ('Estimate', 'customfield_12316544')},
```


sync2jira/downstream_issue.py
```diff
@@ -947,6 +954,24 @@ def _update_jira_labels(issue, labels):
...
+def _update_github_project_fields(client, existing, issue, github_project_fields):
+    """Update a Jira issue with GitHub project item field values
+
+    :param jira.client.JIRA client: JIRA client
+    :param jira.resource.Issue existing: Existing JIRA issue
+    :param sync2jira.intermediary.Issue issue: Upstream issue
+    :param list: Fields representing GitHub project item fields in GitHub and Jira
+    """
+    github_project_fields = issue.downstream.get('github_project_fields', {})
+
```
webbnh, line 966: This assignment overwrites the parameter value.  Assuming that we fetched the appropriate value at line 738 and passed it into this function, we can omit this operation from here.

However, note that the fallback value at line 738 is an empty list while the fallback value here is an empty dictionary.  (I think we want the fallback from here but we want to make the call at line 738.)

sync2jira/upstream_issue.py
```diff
@@ -358,6 +361,25 @@ def github_issues(upstream, config):
         if issue.get('milestone', None):
             issue['milestone'] = issue['milestone']['title']
 
+        if not issue.get('storypoints', None):
+            issue['storypoints'] = ''
+            graphqlurl = 'https://api.github.com/graphql'
+            orgname = "open-connectors"
+            reponame = "open-connectors"
```
webbnh, line 368: `graphqlurl` should probably be defined statically (e.g., as a module-level definition or in some sort of external configuration).

`orgname` and `reponame` should be defined externally.

sync2jira/upstream_issue.py
```diff
@@ -358,6 +361,25 @@ def github_issues(upstream, config):
...
+            issuenumber = issue['number']
+            github_project_fields = config['sync2jira']['map']['github']["/".join((orgname,reponame))]['github_project_fields']
+            for fieldname, values in github_project_fields.items():
+                ghfieldname = values[0]
+                query = """query MyQuery($orgname: String!, $reponame: String!, $ghfieldname: String!, $issuenumber: Int!) { repository(owner: $orgname, name: $reponame) { issue(number: $issuenumber) { title body projectItems(first: 1) { nodes { fieldValueByName(name: $ghfieldname) { ... on ProjectV2ItemFieldNumberValue { number } } } } } }}"""
+                variables = {"orgname": orgname, "reponame": reponame, "ghfieldname": ghfieldname, "issuenumber": issuenumber}
```
webbnh, line 374: `query` should probably be defined statically.  And, unless you're going to use line breaks and pretty formatting, you should use single-double-quotes rather than triple-double-quotes.

Also, since only `ghfieldname` changes inside the loop, consider moving the definition of `variables` outside the loop (possibly inlining and removing the local variables, `orgname`, `reponame`, and `issuenumber`), and then replacing line 374 with something like
```python
                variables["ghfieldname"] = ghfieldname
```

webbnh (approved):
Baiju, here are some things that I noted in a quick scan.

sync2jira/downstream_issue.py
baijum: It was a regression that happened after changing the data structure. Fixed in 46248b8bbea26b8508001cedd23126f6719599b9

sync2jira/upstream_issue.py
baijum: Done. 46248b8bbea26b8508001cedd23126f6719599b9

sync2jira/upstream_issue.py
baijum: Done. 46248b8bbea26b8508001cedd23126f6719599b9

sync2jira/upstream_issue.py
```diff
@@ -34,6 +34,25 @@
...
+                    fieldValueByName(name: $ghfieldname) {
+                        ... on ProjectV2ItemFieldNumberValue {
+                            number
+                        }
+                    }
+                }
+            }
+        }
+    }
+}'''
```
webbnh, line 55: The formatting here seems slightly strange.  If GraphQL really doesn't care about whitespace, I recommend something like this:
```python
project_items_query = '''
    query MyQuery(
        $orgname: String!, $reponame: String!, $ghfieldname: String!, $issuenumber: Int!
    ) {
        repository(owner: $orgname, name: $reponame) {
            issue(number: $issuenumber) {
                title
                body
                projectItems(first: 1) {
                    nodes {
                        fieldValueByName(name: $ghfieldname) {
                            ... on ProjectV2ItemFieldNumberValue {
                                number
                            }
                        }
                    }
                }
            }
        }
    }
'''
```

sync2jira/upstream_issue.py
```diff
@@ -363,16 +382,13 @@ def github_issues(upstream, config):
 
         if not issue.get('storypoints', None):
             issue['storypoints'] = ''
-            graphqlurl = 'https://api.github.com/graphql'
-            orgname = "open-connectors"
-            reponame = "open-connectors"
+            orgname, reponame = upstream.split('/')
             issuenumber = issue['number']
```
webbnh, line 386: Since you intend for `split()` to produce exactly two results, consider specifying [the second argument](https://docs.python.org/3/library/stdtypes.html#str.split):
```python
            orgname, reponame = upstream.split('/', 1)
```

webbnh: Also, consider using [rsplit()](https://docs.python.org/3/library/stdtypes.html#str.rsplit) instead.

(It makes no difference unless and until `upstream` shows up with more than one `/` in it, which we probably hope never happens, but....)

baijum: Done. 1afad545ee80433015e30fc217161ccda367d316

sync2jira/upstream_issue.py
baijum: Done. 1afad545ee80433015e30fc217161ccda367d316

webbnh (approved):
LGTM

mattreid:
We talked about it in stand up today, but I think it would be helpful to consider a default mapping, which could be set on our instance, so that every team syncing story points doesn't have to look up and know the custom field it maps to in our Jira. It'd be nicer for them to just indicate that they want to sync their story point estimates and sync2jira knows how to map that for them.

sync2jira/downstream_issue.py
```diff
@@ -734,6 +734,13 @@ def _update_jira_issue(existing, issue, client):
     if not updates:
         return
 
+    # Get fields representing project item fields in GitHub and Jira
+    github_project_fields = issue.downstream.get('github_project_fields', {})
```
ralphbean, line 738: I'm confused. Is `github_project_fields` a field in JIRA?

ralphbean: Oh, I see. It is part of the `map`.

sync2jira/downstream_issue.py
```diff
@@ -947,6 +954,23 @@ def _update_jira_labels(issue, labels):
...
+    :param jira.client.JIRA client: JIRA client
+    :param jira.resource.Issue existing: Existing JIRA issue
+    :param sync2jira.intermediary.Issue issue: Upstream issue
+    :param list: Fields representing GitHub project item fields in GitHub and Jira
+    """
+
+    for fieldname, values in github_project_fields.items():
+        try:
+            if fieldname == 'storypoints':
+                existing.update({values[1]: str(issue.storypoints)})
```
ralphbean, line 969: Do we need to hardcode a conditional for `'storypoints'` here?

```suggestion
    for name, values in github_project_fields.items():
        try:
            ghfieldname, jirafieldname = values
            existing.update({jirafieldname: str(getattr(issue, name))})
```

sync2jira/upstream_issue.py
```diff
@@ -358,6 +383,22 @@ def github_issues(upstream, config):
...
             issue['milestone'] = issue['milestone']['title']
 
+        if not issue.get('storypoints', None):
+            issue['storypoints'] = ''
+            orgname, reponame = upstream.rsplit('/', 1)
+            issuenumber = issue['number']
+            github_project_fields = config['sync2jira']['map']['github'][upstream]['github_project_fields']
+            variables = {"orgname": orgname, "reponame": reponame, "issuenumber": issuenumber}
+            for fieldname, values in github_project_fields.items():
+                variables['ghfieldname'] = values[0]
```
ralphbean, line 393: I don't know how much value this adds, but writing it this way might help future readers understand what the structure of that tuple is supposed to be more easily.

```suggestion
                ghfieldname, jirafieldname = values
                variables['ghfieldname'] = ghfieldname
```

sync2jira/upstream_issue.py
```diff
@@ -358,6 +383,22 @@ def github_issues(upstream, config):
...
+                response = requests.post(graphqlurl, headers=headers, json={"query": project_items_query, "variables": variables})
+                data = response.json()
+                if fieldname == 'storypoints':
+                    try:
+                        issue[fieldname] = data['data']['repository']['issue']['projectItems']['nodes'][0]['fieldValueByName']['number']
+                    except (TypeError, KeyError):
+                        continue
```
ralphbean, line 400: We definitely need a log statement here (warning, or error, or.. something else?)

There are so many key-accesses and that one zero-index access that this will surely fail at some point. It will be good to have a clue as to why in the logs.

sync2jira/downstream_issue.py
baijum: Done. b921b55434d2599f3c7462d25f84c52fde66f305

sync2jira/upstream_issue.py
baijum: I have adapted the suggestion from @ralphbean with a small change. b921b55434d2599f3c7462d25f84c52fde66f305 Ignored the second variable as it is not used anywhere.

sync2jira/upstream_issue.py
baijum: Done. b921b55434d2599f3c7462d25f84c52fde66f305

baijum:
> We talked about it in stand up today, but I think it would be helpful to consider a default mapping, which could be set on our instance, so that every team syncing story points doesn't have to look up and know the custom field it maps to in our Jira (removing the need for each team to specify 'customfield_12316544' in their repo config). It'd be nicer for them to just indicate that they want to sync their story point estimates and sync2jira knows how to map that for them.

I have addressed this:  9d5a82fbb4774b2aca07f04871e66a2b7c3bc1bc

sync2jira/upstream_issue.py
```diff
@@ -358,6 +383,26 @@ def github_issues(upstream, config):
...
+            default_github_project_fields = config['sync2jira']['default_github_project_fields']
+            project_github_project_fields = config['sync2jira']['map']['github'][upstream]['github_project_fields']
+            github_project_fields = default_github_project_fields | project_github_project_fields
+            variables = {"orgname": orgname, "reponame": reponame, "issuenumber": issuenumber}
+            for fieldname, values in github_project_fields.items():
+                ghfieldname, _ = values
+                variables['ghfieldname'] = ghfieldname
+                response = requests.post(graphqlurl, headers=headers, json={"query": project_items_query, "variables": variables})
+                data = response.json()
+                if fieldname == 'storypoints':
```
webbnh, line 399: Since this block only cares about `'storypoints'` (based on the `if` tests at lines 386 and 399) I think the loop at line 394 is unnecessary.  I believe you can replace the line with
```python
           fieldname, values = 'storypoints', github_project_fields['storypoints']
```
(or just assign `values` and use `'storypoints'` in place of `fieldname` in the loop body code).

On the other hand, if you want this code to be generalized to handle other fields in addition to `'storypoints'`, then you need to remove the conditional at line 399 and replace lines 386-394 with something like
```python
        github_project_fields = default_github_project_fields | github_config[upstream]['github_project_fields']
        if any(issue.get(f) for f in github_project_fields):
            orgname, reponame = upstream.rsplit('/', 1)
            variables = {"orgname": orgname, "reponame": reponame, "issuenumber": issue['number']}
            for fieldname, values in github_project_fields.items():
```
after hoisting the assignment of `default_github_project_fields` out of the loop at line 333 (since it doesn't change) and adding this definition next to it:
```python
        github_config = config['sync2jira']['map']['github']
```

And, you might want to replace the loop at line 394 with something like this
```python
            for fieldname, values in ((f, v) for f, v in github_project_fields.items() if issue.get(f)):
```
to avoid fields which aren't in `issue`.

(And, if you are going to generalize the code here, you probably want to do something similar at lines 292-293.)

webbnh: However, one thing to note or think about:  my code suggestion above probably won't handle the case where the upstream issue is updated to change a field from non-blank to blank/empty -- my code will not explicitly propagate that change to the downstream...I don't know whether that leaves the downstream unchanged or whether it will be implicitly cleared.

webbnh (commented):
Various comments.

sync2jira/downstream_issue.py
```diff
@@ -963,10 +963,10 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
     """
 
-    for fieldname, values in github_project_fields.items():
+    for name, values in github_project_fields.items():
         try:
-            if fieldname == 'storypoints':
-                existing.update({values[1]: str(issue.storypoints)})
+            _, jirafieldname = values
+            existing.update({jirafieldname: str(getattr(issue, name))})
         except JIRAError:
```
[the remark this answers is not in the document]
baijum: Done. 829fc7063abece8ec99041b666ea63d1f647f748

sync2jira/upstream_issue.py
```diff
@@ -387,16 +387,20 @@ def github_issues(upstream, config):
...
+                ghfieldname, _ = values
+                variables['ghfieldname'] = ghfieldname
                 response = requests.post(graphqlurl, headers=headers, json={"query": project_items_query, "variables": variables})
                 data = response.json()
                 if fieldname == 'storypoints':
                     try:
                         issue[fieldname] = data['data']['repository']['issue']['projectItems']['nodes'][0]['fieldValueByName']['number']
-                    except (TypeError, KeyError):
+                    except (TypeError, KeyError) as err:
+                        log.debug("Error setting GitHub field. %s", err)
```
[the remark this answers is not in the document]
baijum: Done. 829fc7063abece8ec99041b666ea63d1f647f748

sync2jira/upstream_issue.py
webbnh: OK; hopefully this will all fall out when you add 'priority'.

Zyzyx:
Is the functionality here allowing teams to sync story points in GH to a user-specified field in Jira? i.e. teams can pick the field they want to sync to?

webbnh:
> Is the functionality here allowing teams to sync story points in GH to a user-specified field in Jira? i.e. teams can pick the field they want to sync to?

I expect that it _could_ be used that way, but it is intended to target the field which is part of the RH conventional project schema (which, in _vanilla_ Jira terms is a "custom" field).

mattreid:
> Is the functionality here allowing teams to sync story points in GH to a user-specified field in Jira? i.e. teams can pick the field they want to sync to?

My understanding of where we landed in this iteration is to add a default value which would apply for our teams to save them all from having to know the customfield mapping we expect their estimates to sync into in our instance (seen in line 56 [here](https://github.com/release-engineering/Sync2Jira/pull/203/files#diff-2cc6246c9e7b2a7019c361cc993b64493eca2e3c9a310823dbe22af866ffa6a4R56)          
`'default_github_project_fields': {'storypoints': ('Estimate', 'customfield_12316544')},`

but each team when they create their config in sync2jira for a repo could specify a different Jira field to sync their Estimates to, if desired. They can just say they want to sync storypoints from GH when setting up their repo, and syncjira would know where to put that metadata in Jira.

Zyzyx:
**Story Points** on issues.redhat.com is **customfield_12310243**. 

**customfield_12316544** (which is in this PR) is **Blocked Reason** which is not correct. If this was tested on stage-jira, that might explain the discrepancy though.

Zyzyx:
Expounding a little bit before getting lunch.

I understand the upstream use case here, sync2jira needs to be flexible because each Jira instance is going to have a different custom ID for Story Points. Providing an out-of-the-box default makes sense and apart from changing the ID to match issues.redhat.com I'm fine with merging this PR. Either that or putting a dummy value with an explanation in the comments would work.

Inside the company, I want to be mindful of the standards we have though. Just because Sync2Jira can be customized doesn't mean it should. We standardized on the **Story Points** field for pointing a while back, and I think our hosted instance should continue to enforce that, otherwise the organizational reports are going to lose accuracy and precision.

You guys agree? (Matt/Webb/Ralph in particular)
Any concerns?

Zyzyx:
I merged another PR and introduced conflict here. I gave the GH UI a shot to resolve them, hopefully I didn't mess anything up.

mattreid:
> Expounding a little bit before getting lunch.
> 
> I understand the upstream use case here, sync2jira needs to be flexible because each Jira instance is going to have a different custom ID for Story Points. Providing an out-of-the-box default makes sense and apart from changing the ID to match issues.redhat.com I'm fine with merging this PR. Either that or putting a dummy value with an explanation in the comments would work.
> 
> Inside the company, I want to be mindful of the standards we have though. Just because Sync2Jira can be customized doesn't mean it should. We standardized on the **Story Points** field for pointing a while back, and I think our hosted instance should continue to enforce that, otherwise the organizational reports are going to lose accuracy and precision.
> 
> You guys agree? (Matt/Webb/Ralph in particular) Any concerns?

I agree that internally we want consistency, and the tools we use should help teams be consistent when there's a standard. 

My only concern is making sure we know how much effort should be spent to support that now, in this issue, vs coming back to later as we tackle more fields and have a better idea how to add structure around managing mappings for them up and downstream.  There are other control / consistency levers in place today: repo configs are managed as code, are peer reviewed, and there are org wide dashboards that the resulting Jira data is aggregated from, expecting those standard fields, so mismatches should be visible when a team sees no sizing data is being recognized for their issues. 

Do you want to see a reduction in flexibility before merging this, so that estimate data from GH can only be synced using the provided Jira field value for that instance, and a different Jira field can't be specified by teams when configuring repos? @baijum has done some exploration on syncing priority values, and we could include this requirement with that if it could wait a little while. Teams would need to establish how their GH values map to the priority values in their Jira project. 

> **Story Points** on issues.redhat.com is **customfield_12310243**.
> 
> **customfield_12316544** (which is in this PR) is **Blocked Reason** which is not correct. If this was tested on stage-jira, that might explain the discrepancy though.

Good catch. For the future, do you know of anything that maintains those mappings across stage and prod to see what equivalent fields are? or is that best done by poking APIs and verifying?

webbnh:
> **Story Points** on issues.redhat.com is **customfield_12310243**.
> 
> **customfield_12316544** (which is in this PR) is **Blocked Reason**

Can someone confirm that the staging instance of Jira is different from the production version?  If so, that seems like something that we should push to get corrected!  😬

webbnh:
> Any concerns?

My only concern is the perfect being the enemy of the good -- if what we have here is "good enough" (which, it won't be until it's referencing the right target field...), then we should merge it and iterate.

PR #204, opened 2024-09-24
Doc delinting

webbnh:
Reading through the _Sync2Jira_ docs, I noticed a few typos/misspellings, and I noted a couple of formatting glitches, and then, when I checked out a branch, my IDE pointed out a few more....  This PR endeavors to fix the typos that I found, and, where my changes hit long lines, I wrapped them to 80 columns to make future reviews easier (but, I left unchanged lines unchanged).

Also, in a separate commit, I removed the **Branches** section from the `README.md`:  it looks completely out of date, and I figured that removing it was better than trying to fix it.  Let me know if you want it restored and/or amended.


ralphbean:
All good changes. Thank you!

PR #207, opened 2024-09-27
Refactoring upstream code to DRY it out a bit

webbnh:
This PR is a series of refactorings and other modest changes to the `upstream_*.py` modules.  The intention is that there be no overall functional change; however, internally, the changes remove a bunch of code duplication, shedding a little over 25% of the previous code.

Each of the first several commits is targeted at removing a particular code duplication.  The commits after that are more code-quality/stylistic changes.

I recommend reviewing it one commit at a time.

A couple of notes on the tests:
- The unit tests, via the mocking, have slightly more insight into the internal structure of the code than I might have wished, so I was not able to use them to validate my changes without also making some changes to them (which kind of undermines their purpose...); however, the test changes were fairly limited (the name of the function being mocked, the change to pass its second parameter by keyword-argument, and the module-path to where the `requests` package is to be patched), so I think the unit tests still did their job.
- At this point, this PR is not expected to be merged as is -- it is mostly intended to record this work for future posterity -- so, there are no new tests for the new subroutines.  Before merging this PR, new tests should be added.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+arrow
 jira
 requests
 requests_kerberos
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -109,8 +109,8 @@ def get_jira_client(issue, config):
 
     if not isinstance(issue, Issue) and not isinstance(issue, PR):
         log.error("passed in issue is not an Issue instance")
-        log.error("It is a %s" % type(issue).__name__)
-        raise TypeError("Got %s, expected Issue" % type(issue).__name__)
+        log.error("It is a %s", type(issue).__name__)
+        raise TypeError(f"Got {type(issue).__name__}, expected Issue")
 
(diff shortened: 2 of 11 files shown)


Zyzyx:
I submitted 2 PRs that reclaimed value for the last few commits (the styling). We should get the first 5 too, but they don't include the story point changes, so that's been tying me up.

webbnh:
@Zyzyx, I've rebased this branch on the branch in #219 which is on the end of `main`, and I've confirmed that the unit tests run successfully on all the remaining commits.  Thus, the commits from #219 also show up here, but the commits that you've cherry-picked from here and put on `main` are gone from here, as are the changes made for #208.

I _think_ I've properly incorporated the Story Point work from #203, but I need to check my work.

webbnh:
OK, it looks good.

webbnh:
This branch was merged as a series of individual PRs.

PR #208, opened 2024-09-27
Remove Pagure Support

Zyzyx:


sync2jira/main.py
```diff
@@ -128,7 +114,7 @@ def load_config(loader=fedmsg.config.load_config):
     if 'map' not in config['sync2jira']:
         raise ValueError("No sync2jira.map section found in fedmsg.d/ config")
 
-    possible = set(['pagure', 'github'])
+    possible = set(['github'])
```
webbnh, line 117: Consider using a "set literal" instead:
```suggestion
    possible = {'github'}
```

sync2jira/main.py
```diff
@@ -152,7 +138,7 @@ def load_config(loader=fedmsg.config.load_config):
 
 def listen(config):
     """
-    Listens to activity on upstream repos on pagure and github \
+    Listens to activity on upstream repos on github \
     via fedmsg, and syncs new issues there to the JIRA instance \
     defined in 'fedmsg.d/sync2jira.py'
```
webbnh, line 143: GitHub should, generally speaking, be spelled with two capital letters (at least in English text).

Also, we shouldn't need the line continuation (backslash) characters in a docstring.

sync2jira/main.py
```diff
@@ -299,7 +262,7 @@ def initialize_recent(config):
     :return: Nothing
     """
     # Query datagrepper
-    ret = query(category=['github', 'pagure'], delta=int(600), rows_per_page=100)
+    ret = query(category=['github'], delta=int(600), rows_per_page=100)
```
webbnh, line 265: Off topic, I admit (my apologies...), but it strikes me that, given modern machines and networking, 100 rows per page seems pretty small...especially if _Sync2Jira_ is about to get a whole lot more traffic....

tests/test_main.py
```diff
@@ -83,32 +82,25 @@ def test_close_duplicates(self,
         """
         # Set up return values
         mock_load_config.return_value = self.mock_config
-        mock_u.pagure_issues.return_value = ['mock_issue_github']
-        mock_u.github_issues.return_value = ['mock_issue_pagure']
+        mock_u.github_issues.return_value = ['mock_issue_github']
```
webbnh, line 85: Were the return values in the old code _supposed_ to be reversed, or was that a bug that you've corrected here??

tests/test_main.py
```diff
@@ -284,8 +248,7 @@ def test_initialize_errors(self,
             m.initialize_issues(self.mock_config)
 
         # Assert everything was called correctly
-        mock_u.pagure_issues.assert_called_with('key_pagure', self.mock_config)
-        mock_d.sync_with_jira.assert_any_call('mock_issue_pagure', self.mock_config)
+        mock_u.github_issues.assert_called_with('key_github', self.mock_config)
```
webbnh, line 251: You don't want an assertion here like this?
```python
        mock_d.sync_with_jira.assert_any_call('mock_issue_github', self.mock_config)
```

tests/test_main.py
```diff
@@ -301,15 +264,12 @@ def test_initialize_api_limit(self,
...
-        mock_u.pagure_issues.return_value = ['mock_issue_pagure']
         mock_u.github_issues.side_effect = mock_error
 
         # Call the function
         m.initialize_issues(self.mock_config, testing=True)
 
         # Assert everything was called correctly
-        mock_u.pagure_issues.assert_called_with('key_pagure', self.mock_config)
-        mock_d.sync_with_jira.assert_any_call('mock_issue_pagure', self.mock_config)
         mock_u.github_issues.assert_called_with('key_github', self.mock_config)
```
webbnh, line 273: Do you want to change (old) line 312 to
```python
        mock_d.sync_with_jira.assert_any_call('mock_issue_github', self.mock_config)
```
or does the exception prevent the call (in which case, should we be asserting that it does _not_ get called?)?

Ditto for `test_initialize_github_error()`.

webbnh (approved):
I only reviewed the changes that you made (I didn't try to find other changes that you _should_ have made...), but what's here looks good to me.

It looks like you found and fixed a test bug!  That's awesome.  I picked out a couple of other possibilities for you to consider.

FWIW, I heartily support the changes you made to put the function call parameters all on one line where they fit!  And, my IDE was complaining about some of the indentation, too, so I'm glad you fixed that.

In addition, I called out a couple of other somewhat off-topic nits, and I made one other comment, just something to put in the back of your mind.

sync2jira/main.py
Zyzyx: I agree, but I think that change is out of scope for this PR.

webbnh: Agreed.

tests/test_main.py
Zyzyx: I didn't test the old code to confirm, but I did notice that and assumed it was a oversight.

sync2jira/main.py
Zyzyx: I'll take care of this

tests/test_main.py
Zyzyx: Ah yes, my mistake.

sync2jira/main.py
Zyzyx: see [40a3c50](https://github.com/release-engineering/Sync2Jira/pull/208/commits/40a3c505c426a057771a28ec526f832a0e3d663f)

tests/test_main.py
Zyzyx: Handled in [40a3c50](https://github.com/release-engineering/Sync2Jira/pull/208/commits/40a3c505c426a057771a28ec526f832a0e3d663f)

PR #216, opened 2024-10-04
Sync priority from GitHub to Jira

### LINKED ISSUE #198: Sync Priority values
If priority exists on the synced GH issue, it should apply that to the created Jira issue for better backlog transparency.

baijum:
Fixes #198

Example configuration:
```
"default_jira_fields": {
  "storypoints": "customfield_12310243"
},

"map": {
  "github": {
    "developerproductivity/open-connectors": {
      "project": "TEST",
      "sync": [ "issue" ],
      "issue_updates": [ "github_project_fields" ],
      "github_project_number": 1,
      "github_project_fields": {
        "storypoints": {
          "gh_field": "Estimate"
        },
        "priority": {
          "gh_field": "Priority",
          "options": {
            "P0": "Blocker",
            "P1": "Critical",
            "P2": "Major",
            "P3": "Minor",
            "P4": "Optional",
            "P5": "Trivial"
          }
        }
      }
    }
  }
},
```

### DIFF
--- a/docs/source/config-file.rst
+++ b/docs/source/config-file.rst
@@ -63,6 +63,22 @@ The config file is made up of multiple parts
 
 .. code-block:: python
 
+    "default_jira_fields": {
+      "storypoints": "customfield_12310243"
+    },
+
+* The keys in the :code:`default_jira_fields` map are used as the values of the github_project_fields
+  list (see below). The keys indicate how the fields in GitHub Project items correspond to
+  fields in Jira issues so that these values can be synced properly. Currently, we support
+  only :code:`storypoints` and :code:`priority`. By default, the value for the priority is sourced
+  from the upstream :code:`priority` field, but providing a :code:`priority` key for
+  :code:`default_jira_fields` will override that. Unfortunately, there is no default available
+  for sourcing story points, so, the :code:`storypoints` key must be provided if projects wish
+  to sync story point values, and, for our corporate Jira instance, the value should be specified
+  as :code:`customfield_12310243`.
+
+.. code-block:: python
+
     'map': {
             'github': {
                 'GITHUB_USERNAME/Demo_project': {'project': 'FACTORY', 'component': 'gitbz',
@@ -124,6 +140,10 @@ The config file is made up of multiple parts
         * If selected this will add a comment to all newly created JIRA issue in the format 'UPSTREAM_PROJECT-#1' where the number indicates the issue ID. This allows users to search for the issue on JIRA via the issue number.
     * :code:`url`
         * This flag will add the upstream url to the bottom of the JIRA ticket
+    * :code:`github_project_number`
+        * Specify the GitHub project number. If specified, story points and priority will be selected from this project, and other projects linked to the issues will be ignored.
+    * :code:`github_project_fields`
+        * Sync GitHub projects fields. e.g, storypoints, priority
 
     .. note::
 
@@ -145,6 +165,33 @@ The config file is made up of multiple parts
 
 * It is strongly encouraged for teams to use the :code:`owner` field. If configured, owners will be alerted if Sync2Jira finds duplicate downstream issues.
   Further the owner will be used as a default in case the program is unable to find a valid assignee.
+
+.. code-block:: python
+
+    'github_project_fields': {
+        'storypoints': {
+          'gh_field': 'Estimate'
+        },
+        'priority': {
+          'gh_field': 'Priority',
+          'options': {
+            'P0': 'Blocker',
+            'P1': 'Critical',
+            'P2': 'Major',
+            'P3': 'Minor',
+            'P4': 'Optional',
+            'P5': 'Trivial'
+          }
+        }
+    }
+
+* Specify the GitHub project field and its mapping to Jira. The currently supportted fields are :code:`storypoints`
+  and :code:`priority`.
+
+* The :code:`gh_field` points to the GitHub field name.
+
+* The :code:`options` key is used to specify a mapping between GitHub field values as Jira field values.
+  This is particularly useful for cases like priority which uses keywords for values.
 
 .. code-block:: python
 
--- a/fedmsg.d/sync2jira.py
+++ b/fedmsg.d/sync2jira.py
@@ -53,11 +53,29 @@ config = {
                 'token_auth': 'YOUR_JIRA_ACCESS_TOKEN',
             },
         },
-        'default_github_project_fields': {'storypoints': ('Estimate', 'customfield_12310243')},
+        'default_jira_fields': {
+            'storypoints': 'customfield_12310243',
+            },
         'map': {
             'github': {
                 'GITHUB_USERNAME/Demo_project': {'project': 'FACTORY', 'component': 'gitbz',
-                                                 'updates': [...], 'sync': ['pullrequest', 'issue']},
+                                                'issue_updates': [
+                                                    'comments',
+                                                    'upstream_id',
+                                                    'title',
+                                                    'description',
+                                                    'github_markdown',
+                                                    'upstream_id',
+                                                    'url',
+                                                    {'transition': 'Closed'},
+                                                    {'assignee': {'overwrite': False}},
+                                                    'github_project_fields'],
+                                                'github_project_number': '1',
+                                                'github_project_fields': {'storypoints': {'gh_field': 'Estimate'},
+                                                    'priority': {'gh_field': 'Priority', 'options':
+                                                                 {'P0': 'Blocker', 'P1': 'Critical', 'P2': 'Major',
+                                                                  'P3': 'Minor', 'P4': 'Optional', 'P5': 'Trivial'}}},
+                                                 'sync': ['pullrequest', 'issue']},
             },
         },
         'filters': {
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -703,7 +703,7 @@ def _create_jira_issue(client, issue, config):
 
     # Update relevant information (i.e. tags, assignees etc.) if the
     # User opted in
-    _update_jira_issue(downstream, issue, client)
+    _update_jira_issue(downstream, issue, client, config)
 
     return downstream
 
@@ -725,7 +725,7 @@ def _label_matching(jira_labels, issue_labels):
     return updated_labels
 
 
-def _update_jira_issue(existing, issue, client):
+def _update_jira_issue(existing, issue, client, config):
     """
     Updates an existing JIRA issue (i.e. tags, assignee, comments etc).
 
@@ -751,7 +751,8 @@ def _update_jira_issue(existing, issue, client):
     # Only synchronize comments for listings that op-in
     if 'github_project_fields' in updates and len(github_project_fields) > 0:
         log.info("Looking for GitHub project fields")
-        _update_github_project_fields(client, existing, issue, github_project_fields)
+        _update_github_project_fields(client, existing, issue,
+                                      github_project_fields, config)
 
     # Only synchronize comments for listings that op-in
     if 'comments' in updates:
@@ -967,7 +968,8 @@ def _update_jira_labels(issue, labels):
     log.info('Updated %s tag(s)' % len(_labels))
 
 
-def _update_github_project_fields(client, existing, issue, github_project_fields):
+def _update_github_project_fields(client, existing, issue,
+                                  github_project_fields, config):
     """Update a Jira issue with GitHub project item field values
 
     :param jira.client.JIRA client: JIRA client
@@ -976,13 +978,32 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
     :param list: Fields representing GitHub project item fields in GitHub and Jira
     """
 
+    default_jira_fields = config['sync2jira'].get('default_jira_fields', {})
     for name, values in github_project_fields.items():
-        _, jirafieldname = values
-        try:
-            existing.update({jirafieldname: str(getattr(issue, name))})
-        except JIRAError as err:
-            # Add a comment to indicate there was an issue
-            client.add_comment(existing, f"Error updating GitHub project field: {err}")
+        fieldvalue = getattr(issue, name)
+        if name == 'storypoints':
+            try:
+                jirafieldname = default_jira_fields['storypoints']
+            except KeyError:
+                log.error("Configuration error: Missing 'storypoints' in `default_jira_fields`")
+                continue
+            try:
+                existing.update({jirafieldname: fieldvalue})
+            except JIRAError as err:
+                # Add a comment to indicate there was an issue
+                client.add_comment(existing, f"Error updating GitHub project storypoints field: {err}")
+        elif name == 'priority':
+            try:
+                jirafieldname = default_jira_fields['priority']
+            except KeyError:
+                jirafieldname = 'priority'
+            jira_priority = values.get('options', {}).get(fieldvalue)
+            if jira_priority:
+                try:
+                    existing.update({jirafieldname: {'name': jira_priority}})
+                except JIRAError as err:
+                    # Add a comment to indicate there was an issue
+                    client.add_comment(existing, f"Error updating GitHub project priority field: {err}")
 
 
 def _update_tags(updates, existing, issue):
@@ -1157,7 +1178,7 @@ def sync_with_jira(issue, config):
             log.info("Testing flag is true.  Skipping actual update.")
             return
         # Update relevant metadata (i.e. tags, assignee, etc)
-        _update_jira_issue(existing, issue, client)
+        _update_jira_issue(existing, issue, client, config)
         return
 
     # If we're *not* configured to do legacy matching (upgrade mode) then there
--- a/sync2jira/intermediary.py
+++ b/sync2jira/intermediary.py
@@ -93,7 +93,6 @@ class Issue(object):
         if any('fixVersion' in item for item in mapping):
             map_fixVersion(mapping, issue)
 
-        # TODO: Priority is broken
         return cls(
             source=upstream_source,
             title=issue['title'],
@@ -103,7 +102,7 @@ class Issue(object):
             comments=comments,
             tags=issue['labels'],
             fixVersion=[issue['milestone']],
-            priority=None,
+            priority=issue['priority'],
             content=issue['body'] or '',
             reporter=issue['user'],
             assignee=issue['assignees'],
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -28,26 +28,85 @@ import sync2jira.intermediary as i
 
 log = logging.getLogger('sync2jira')
 graphqlurl = 'https://api.github.com/graphql'
-project_items_query = '''
+
+ghquery = '''
     query MyQuery(
-        $orgname: String!, $reponame: String!, $ghfieldname: String!, $issuenumber: Int!
+        $orgname: String!, $reponame: String!, $issuenumber: Int!
     ) {
         repository(owner: $orgname, name: $reponame) {
-            issue(number: $issuenumber) {
-                title
-                body
-                projectItems(first: 1) {
-                    nodes {
-                        fieldValueByName(name: $ghfieldname) {
-                            ... on ProjectV2ItemFieldNumberValue {
-                                number
-                            }
-                        }
-                    }
+          issue(number: $issuenumber) {
+            title
+            body
+            projectItems(first: 3) {
+              nodes {
+                project {
+                  title
+                  number
+                  url
                 }
+                fieldValues(first: 100) {
+                  nodes {
+                    ... on ProjectV2ItemFieldSingleSelectValue {
+                      name
+                      fieldName: field {
+                        ... on ProjectV2FieldCommon {
+                          name
+                        }
+                      }
+                    }
+                    ... on ProjectV2ItemFieldTextValue {
+                      text
+                      fieldName: field {
+                        ... on ProjectV2FieldCommon {
+                          name
+                        }
+                      }
+                    }
+                    ... on ProjectV2ItemFieldNumberValue {
+                      number
+                      fieldName: field {
+                        ... on ProjectV2FieldCommon {
+                          name
+                        }
+                      }
+                    }
+                    ... on ProjectV2ItemFieldDateValue {
+                      date
+                      fieldName: field {
+                        ... on ProjectV2FieldCommon {
+                          name
+                        }
+                      }
+                    }
+                    ... on ProjectV2ItemFieldUserValue {
+                      users(first: 10) {
+                        nodes {
+                          login
+                        }
+                      }
+                      fieldName: field {
+                        ... on ProjectV2FieldCommon {
+                          name
+                        }
+                      }
+                    }
+                    ... on ProjectV2ItemFieldIterationValue {
+                      title
+                      duration
+                      startDate
+                      fieldName: field {
+                        ... on ProjectV2FieldCommon {
+                          name
+                        }
+                      }
+                    }
+                  }
+                }
+              }
             }
+          }
         }
-    }
+      }
 '''
 
 
@@ -202,8 +261,10 @@ def github_issues(upstream, config):
     # And get comments if needed
     github_client = Github(config['sync2jira']['github_token'], retry=5)
 
+    orgname, reponame = upstream.rsplit('/', 1)
     # We need to format everything to a standard to we can create an issue object
     final_issues = []
+    project_number = config['sync2jira']['map']['github'][upstream].get('github_project_number')
     for issue in issues:
         # Update comments:
         # If there are no comments just make an empty array
@@ -257,26 +318,41 @@ def github_issues(upstream, config):
         if issue.get('milestone', None):
             issue['milestone'] = issue['milestone']['title']
 
-        if not issue.get('storypoints', None):
-            issue['storypoints'] = ''
-            orgname, reponame = upstream.rsplit('/', 1)
+        issue_updates = config['sync2jira']['map']['github'][upstream].get('issue_updates', {})
+
+        if 'github_project_fields' in issue_updates:
+            issue['storypoints'] = None
+            issue['priority'] = ''
             issuenumber = issue['number']
-            default_github_project_fields = config['sync2jira']['default_github_project_fields']
-            project_github_project_fields = config['sync2jira']['map']['github'].get(upstream, {}).get('github_project_fields', {})
-            github_project_fields = default_github_project_fields | project_github_project_fields
+            github_project_fields = config['sync2jira']['map']['github'][upstream]['github_project_fields']
             variables = {"orgname": orgname, "reponame": reponame, "issuenumber": issuenumber}
-            for fieldname, values in github_project_fields.items():
-                ghfieldname, _ = values
-                variables['ghfieldname'] = ghfieldname
-                response = requests.post(graphqlurl, headers=headers, json={"query": project_items_query, "variables": variables})
-                data = response.json()
-                if fieldname == 'storypoints':
-                    try:
-                        issue[fieldname] = data['data']['repository']['issue']['projectItems']['nodes'][0]['fieldValueByName']['number']
-                    except (TypeError, KeyError) as err:
-                        log.debug("Error fetching %s!r from GitHub %s/%s#%s: %s",
-                                  ghfieldname, orgname, reponame, issuenumber, err)
-                        continue
+            response = requests.post(graphqlurl, headers=headers, json={"query": ghquery, "variables": variables})
+            if response.status_code != 200:
+                log.debug("Error while fetching %s/%s/Issue#%s: %s", orgname, reponame, issuenumber, response.text)
+                continue
+            data = response.json()
+            gh_issue = data['data']['repository']['issue']
+            if not gh_issue:
+                log.debug("Error fetching issue from GitHub: %s. org: %s, repo: %s", response.text, orgname, reponame)
+                continue
+            gh_field_name = ''
+            try:
+                currentProjectNode = _get_current_project_node(upstream, project_number, issuenumber, gh_issue)
+                if not currentProjectNode:
+                    continue
+                for item in currentProjectNode['fieldValues']['nodes']:
+                    if 'fieldName' in item:
+                        gh_field_name = item['fieldName'].get('name')
+                        if 'priority' in github_project_fields\
+                                and gh_field_name == github_project_fields['priority']['gh_field']:
+                            issue['priority'] = item.get('name')
+                        if 'storypoints' in github_project_fields\
+                                and gh_field_name == github_project_fields['storypoints']['gh_field']:
+                            issue['storypoints'] = int(item['number'])
+            except KeyError as err:
+                log.debug("Error fetching %s!r from GitHub %s/%s#%s: %s",
+                          gh_field_name, orgname, reponame, issuenumber, err)
+                continue
 
         final_issues.append(issue)
 
@@ -286,6 +362,32 @@ def github_issues(upstream, config):
     ))
     for issue in final_issues:
         yield issue
+
+
+def _get_current_project_node(upstream, project_number, issue_number, gh_issue):
+    project_items = gh_issue['projectItems']['nodes']
+    # If there are no project items, there is nothing to return.
+    if len(project_items) == 0:
+        log.debug("Issue %s/%s is not associated with any project", upstream, issue_number)
+        return None
+    # If there is exactly one project item, return it if we don't have a configured project.
+    if not project_number and len(project_items) == 1:
+        return project_items[0]
+    # There are multiple projects associated with this issue; if we don't have a configured
+    # project, then we don't know which one to return, so return none.
+    if not project_number:
+        prj = ", ".join([":".join([x['project']['url'], x['project']['title']]) for x in project_items])
+        log.debug(
+            "Project number is not configured, and the issue %s/%s is associated with more than one project: %s",
+            upstream, issue_number, prj)
+        return None
+    # Return the associated project which matches the configured project if we can find it.
+    for item in project_items:
+        if item['project']['number'] == project_number:
+            return item
+
+    log.debug("Issue is not associated with the configured project, but other associations exist.")
+    return None
 
 
 def get_all_github_data(url, headers):
--- a/tests/test_downstream_issue.py
+++ b/tests/test_downstream_issue.py
@@ -28,6 +28,7 @@ class TestDownstreamIssue(unittest.TestCase):
             'sync2jira': {
                 'default_jira_instance': 'another_jira_instance',
                 'jira_username': 'mock_user',
+                'default_jira_fields': {'storypoints': 'customfield_12310243'},
                 'jira': {
                     'mock_jira_instance': {'mock_jira': 'mock_jira'},
                     'another_jira_instance': {'token_auth': 'mock_token',
@@ -70,6 +71,8 @@ class TestDownstreamIssue(unittest.TestCase):
         self.mock_issue.assignee = [{'fullname': 'mock_assignee'}]
         self.mock_issue.status = 'Open'
         self.mock_issue.id = '1234'
+        self.mock_issue.storypoints = 2
+        self.mock_issue.priority = 'P1'
 
         # Mock issue updates
         self.mock_updates = [
@@ -128,7 +131,6 @@ class TestDownstreamIssue(unittest.TestCase):
         """
         # Set up return values
         self.mock_issue.downstream = {}
-        self.mock_config['sync2jira']['default_jira_instance'] = {}
 
         # Call the function
         with self.assertRaises(Exception):
@@ -370,7 +372,8 @@ class TestDownstreamIssue(unittest.TestCase):
         mock_update_jira_issue.assert_called_with(
             self.mock_downstream,
             self.mock_issue,
-            mock_client
+            mock_client,
+            self.mock_config
         )
         self.mock_downstream.update.assert_any_call({'customfield_1': 'DUMMY-1234'})
         self.mock_downstream.update.assert_any_call({'customfield_2': 'dummy@dummy.com'})
@@ -424,7 +427,8 @@ class TestDownstreamIssue(unittest.TestCase):
         mock_update_jira_issue.assert_called_with(
             self.mock_downstream,
             self.mock_issue,
-            mock_client
+            mock_client,
+            self.mock_config
         )
         self.mock_downstream.update.assert_any_call({'customfield_1': 'DUMMY-1234'})
         self.mock_downstream.update.assert_any_call(
@@ -479,7 +483,8 @@ class TestDownstreamIssue(unittest.TestCase):
         mock_update_jira_issue.assert_called_with(
             self.mock_downstream,
             self.mock_issue,
-            mock_client
+            mock_client,
+            self.mock_config
         )
         self.mock_downstream.update.assert_any_call({'customfield_1': 'DUMMY-1234'})
         self.mock_downstream.update.assert_any_call(
@@ -533,7 +538,8 @@ class TestDownstreamIssue(unittest.TestCase):
         mock_update_jira_issue.assert_called_with(
             self.mock_downstream,
             self.mock_issue,
-            mock_client
+            mock_client,
+            self.mock_config
         )
         self.assertEqual(response, self.mock_downstream)
         mock_client.add_comment.assert_not_called()
@@ -571,7 +577,8 @@ class TestDownstreamIssue(unittest.TestCase):
 
         # Assert all calls were made correctly
         mock_get_jira_client.assert_called_with(self.mock_issue, self.mock_config)
-        mock_update_jira_issue.assert_called_with(self.mock_downstream, self.mock_issue, mock_client)
+        mock_update_jira_issue.assert_called_with(self.mock_downstream, self.mock_issue,
+                                                  mock_client, self.mock_config)
         mock_create_jira_issue.assert_not_called()
         mock_existing_jira_issue_legacy.assert_not_called()
 
@@ -673,7 +680,8 @@ class TestDownstreamIssue(unittest.TestCase):
         d._update_jira_issue(
             existing=self.mock_downstream,
             issue=self.mock_issue,
-            client=mock_client
+            client=mock_client,
+            config=self.mock_config
         )
 
         # Assert all calls were made correctly
@@ -1665,3 +1673,39 @@ class TestDownstreamIssue(unittest.TestCase):
 
         # Assert everything was called correctly
         self.mock_downstream.update.assert_not_called()
+
+    @mock.patch('jira.client.JIRA')
+    def test_update_github_project_fields_storypoints(self, mock_client):
+        """
+        This function tests `_update_github_project_fields`
+        with story points value.
+        """
+        github_project_fields = {
+         "storypoints": {
+           "gh_field": "Estimate"
+         }}
+        d._update_github_project_fields(mock_client, self.mock_downstream, self.mock_issue,
+                                  github_project_fields, self.mock_config)
+        self.mock_downstream.update.assert_called_with({'customfield_12310243': 2})
+
+
+    @mock.patch('jira.client.JIRA')
+    def test_update_github_project_fields_priority(self, mock_client):
+        """
+        This function tests `_update_github_project_fields`
+        with priority value.
+        """
+        github_project_fields = {
+         "priority": {
+           "gh_field": "Priority",
+           "options": {
+             "P0": "Blocker",
+             "P1": "Critical",
+             "P2": "Major",
+             "P3": "Minor",
+             "P4": "Optional",
+             "P5": "Trivial"
+        }}}
+        d._update_github_project_fields(mock_client, self.mock_downstream, self.mock_issue,
+                                  github_project_fields, self.mock_config)
+        self.mock_downstream.update.assert_called_with({'priority': {'name': 'Critical'}})
--- a/tests/test_intermediary.py
+++ b/tests/test_intermediary.py
@@ -35,6 +35,7 @@ class TestIntermediary(unittest.TestCase):
             'labels': 'mock_tags',
             'milestone': 'mock_milestone',
             'priority': 'mock_priority',
+            'storypoints': '1.0',
             'body': 'mock_content',
             'user': 'mock_reporter',
             'assignees': 'mock_assignee',
@@ -73,7 +74,6 @@ class TestIntermediary(unittest.TestCase):
         self.assertEqual(response.upstream, 'github')
         self.assertEqual(response.comments, [{'body': 'mock_body', 'name': 'mock_name', 'author': 'mock_author',
                                               'changed': None, 'date_created': 'mock_date', 'id': 'mock_id'}])
-        self.assertEqual(response.priority, None)
         self.assertEqual(response.content, 'mock_content')
         self.assertEqual(response.reporter, 'mock_reporter')
         self.assertEqual(response.assignee, 'mock_assignee')
@@ -94,6 +94,7 @@ class TestIntermediary(unittest.TestCase):
         self.checkResponseFields(response)
 
         self.assertEqual(response.fixVersion, ['mock_milestone'])
+        self.assertEqual(response.priority, 'mock_priority')
         self.assertEqual(response.status, 'Open')
         self.assertEqual(response.downstream, {'mock_downstream': 'mock_key'})
         self.assertEqual(response.storypoints, 'mock_storypoints')
@@ -117,6 +118,7 @@ class TestIntermediary(unittest.TestCase):
 
         self.assertEqual(response.tags, 'mock_tags')
         self.assertEqual(response.fixVersion, ['mock_milestone'])
+        self.assertEqual(response.priority, 'mock_priority')
         self.assertEqual(response.status, 'Closed')
         self.assertEqual(response.downstream, {'mock_downstream': 'mock_key'})
         self.assertEqual(response.storypoints, 'mock_storypoints')
@@ -144,6 +146,7 @@ class TestIntermediary(unittest.TestCase):
 
         self.assertEqual(response.tags, 'mock_tags')
         self.assertEqual(response.fixVersion, ['Test mock_milestone'])
+        self.assertEqual(response.priority, 'mock_priority')
         self.assertEqual(response.status, 'Closed')
         self.assertEqual(response.downstream, {
             'mock_downstream': 'mock_key',
--- a/tests/test_upstream_issue.py
+++ b/tests/test_upstream_issue.py
@@ -15,12 +15,9 @@ class TestUpstreamIssue(unittest.TestCase):
     def setUp(self):
         self.mock_config = {
             'sync2jira': {
-                'default_github_project_fields': {
-                    'storypoints': ('Estimate', 'customfield_12310243')
-                },
                 'map': {
                     'github': {
-                        'org/repo': {'sync': ['issue']},
+                        'org/repo': {'sync': ['issue']}
                     },
                 },
                 'jira': {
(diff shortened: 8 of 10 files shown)


kpiwko:
For configuration, what is a valid configuration?

```
'github_project_fields': {
    'storypoints': {
        'fieldmap': {
            'Estimate': 'customfield_12310243'
        }
    },
    'priority': {
        'fieldmap': {
            'Priority': 'priority'
        },
        'options': {
            'P0': 'Blocker',
            'P1': 'Critical',
            'P2': 'Major',
            'P3': 'Minor',
            'P4': 'Optional',
            'P5': 'Trivial'
        }
    }
}
```

Can I have this?
```
'github_project_fields': {
    'storypoints': {
        'fieldmap': {
            'Estimate': 'customfield_12310243',
            'Estimate': 'customfield_123455555',
             'Size': 'customfield_123455'
        }
    },
    'priority': {
        'fieldmap': {
            'Priority': 'priority'
        },
        'options': {
            'P0': 'Blocker',
            'P1': 'Critical',
            'P2': 'Major',
            'P3': 'Minor',
            'P4': 'Optional',
            'P5': 'Trivial'
        }
    }
}
```

If not, the structure can be simplified. What I don't know, is whether there is a use case where two different gh project with different field configurations need to be processed by a single configuration of sync2jira.

kpiwko:
You might want to have something like this as a configuration instead?
```
github_project_field = {
    'gh_project': 'foo/bar' # this might be a regex/wildcard to share configuration?
    'storypoints': {
        'jira_field': 'customfield_12310243',
        'github_field': 'Estimate'
    },
    'priority': {
        'jira_field': 'priority',
        'github_field': 'Priority',
        'value_mappings': {
            'P0': 'Blocker',
            'P1': 'Critical',
            'P2': 'Major',
            'P3': 'Minor',
            'P4': 'Optional',
            'P5': 'Trivial'
        }
    }
}
```

baijum:
> You might want to have something like this as a configuration instead?
> 
> ```
> github_project_field = {
>     'gh_project': 'foo/bar' # this might be a regex/wildcard to share configuration?
>     'storypoints': {
>         'jira_field': 'customfield_12310243',
>         'github_field': 'Estimate'
>     },
>     'priority': {
>         'jira_field': 'priority',
>         'github_field': 'Priority',
>         'value_mappings': {
>             'P0': 'Blocker',
>             'P1': 'Critical',
>             'P2': 'Major',
>             'P3': 'Minor',
>             'P4': 'Optional',
>             'P5': 'Trivial'
>         }
>     }
> }
> ```

This config structure is better. Let's wait for others' comments.

kpiwko:
Thinking about it, `storypoints` and `priority` should be further wrapped in a an object, e.g.:
```
github_project_field_mapping = {
  gh_project: string
  mappings: {
    mapping_name_1: {
       jira_field: string
       github_field: string
       value_mapping: string[] | None
    }
   mapping_name_2: ...
   ....
  }
}
```

sync2jira/downstream_issue.py
```diff
@@ -960,12 +960,24 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
     for name, values in github_project_fields.items():
-        _, jirafieldname = values
-        try:
-            existing.update({jirafieldname: str(getattr(issue, name))})
-        except JIRAError:
-            # Add a comment to indicate there was an issue
-            client.add_comment(existing, f"Error updating GitHub project field")
+        fieldvalue = str(getattr(issue, name))
+        if not fieldvalue:
+            continue
```
webbnh, line 965: I don't think that this is going to do what you want it to:  converting the `getattr()` return to a `str` will produce `"None"` when the attribute is not present, and that value is _not_ "false-y".

Why are you coercing the value to be a string?  I would think it would normally be a string (without conversion), and otherwise you'll trap it with the `if` statement.  So, if you just remove the `str()` conversion, then the code should work.

sync2jira/downstream_issue.py
```diff
@@ -960,12 +960,24 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
+                # Add a comment to indicate there was an issue
+                client.add_comment(existing, f"Error updating GitHub project storypoints field")
+        elif name == 'priority':
+            jira_priority = github_project_fields['priority']['options'][fieldvalue]
+            try:
+                existing.update({'priority': {'name': jira_priority}})
+            except JIRAError:
+                # Add a comment to indicate there was an issue
+                client.add_comment(existing, f"Error updating GitHub project priority field")
+            pass
```
webbnh, line 980: The cascading `if` "waterfall" structure is disappointing.  If we construct the `github_project_fields` data structure correctly, we should be able to do straightforward key-lookups instead of a series of `if` tests.

sync2jira/downstream_issue.py
```diff
@@ -960,12 +960,24 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
-        try:
-            existing.update({jirafieldname: str(getattr(issue, name))})
-        except JIRAError:
-            # Add a comment to indicate there was an issue
-            client.add_comment(existing, f"Error updating GitHub project field")
+        fieldvalue = str(getattr(issue, name))
+        if not fieldvalue:
+            continue
+        if name == 'storypoints':
+            _, jirafieldname = tuple(values['fieldmap'].items())[0]
```
webbnh, line 967: Rather than instantiating the tuple just to grab the first item from it, I suggest using an iterator:
```suggestion
        if name == 'storypoints':
            _, jirafieldname = next(iter(values['fieldmap'].items()))
```
or
```suggestion
        if name == 'storypoints':
            _, jirafieldname = next((k,v) for k, v in values['fieldmap'].items())
```

webbnh (changes requested):
These changes treat story points and priority values separately and individually.  I think they should just be two items in a list of fields to be sync'd, and that the _code_ should not explicitly reference names like `'storypoints'` or `'priority'` -- these should just be keys in a configuration `dict`.

I think that we should come up with a configuration structure which contains all the relevant information (upstream fieldname, downstream fieldname, value-conversion table, etc.) so that the code doesn't have to concern itself with any of the details.

Also, I think we should be able to obtain all the field values for a given issue/PR with a single query to the upstream repo; and, hopefully, we can set them all with a single request to the downstream repo.  So, I think we should be structuring the code toward those ends.

Those opinions aside, below I've called out a bug and a bunch of coding suggestions.
In any case, before you proceed with this PR you need to update the unit tests (and maybe add to them!) and get your code pass Tox.

Note that the current tip of `main` is not currently passing the tests.  😞  I'll be posting a PR shortly which addresses this.

webbnh:
> Thinking about it, `storypoints` and `priority` should be further wrapped in a an object, e.g.:
> 
> ```
> github_project_field_mapping = {array of [
>   gh_project: string
>   mappings: {
>     mapping_name_1: {
>        jira_field: string
>        github_field: string
>        value_mapping: optional Record<string,string> 
>     }
>    mapping_name_2: ...
>    ....
>   ]}
> }
> ```

@kpiwko, I think you're coming into this "off by one level".  I believe that the `github_project_field_mapping` is supposed to provide a list of fields (and details of how they are mapped) which a given project _could_ request to have sync'd.  As such, it should not be including the `gh_project` in this table.  Nevertheless, I do agree that it should be a collection of mappings, although whether that collection should be a list or a dict is another question.

I suggest "backing into" the configuration definition by looking at how it will be used.  I believe that the references will come from the `map` substructure:
```python
config = {
    'sync2jira': {

        ...

        'default_github_project_fields': {'storypoints': ('Estimate', 'customfield_12310243')},
        'map': {
            'github': {
                'GITHUB_USERNAME/Demo_project': {'project': 'FACTORY', 'component': 'gitbz',
                                                 'updates': [...], 'sync': ['pullrequest', 'issue']},
            },
        },
```
and the question is, what goes in the `'updates'` list?  (BTW, I pulled that text from [here](https://github.com/release-engineering/Sync2Jira/blob/83b751d6531598791f7b47d305b64ce6c4f41875/fedmsg.d/sync2jira.py#L57-L62), should `'updates'` in fact be `'issue_updates'`??).  Pulling an example from somewhere else, I see this:
```python
        issue_updates=[
          'comments',
          'description',
          'title',
          'github_markdown',
          {'on_close': {'apply_labels': ['closed-upstream']}},
          {'transition': 'Closed'},
          'url',
        ],
```
and I think we are talking about adding items to this list.  The question is, should those items be the names of the upstream fields or should they be some more abstract "mapping name".  I'm inclined to use the upstream field name for simplicity.  Given that, then the configuration structure should presumably look something like this:
```python
config = {
    'sync2jira': {

        ...

        'default_github_project_fields': {
            'Estimate': {
                'fieldmap': 'customfield_12310243'
            },
            'Priority': {
                'fieldmap': 'priority'
                'options': {  # I like "valuemap" better, in juxtaposition to "fieldmap"
                    'P0': 'Blocker',
                    'P1': 'Critical',
                    'P2': 'Major',
                    'P3': 'Minor',
                    'P4': 'Optional',
                    'P5': 'Trivial'
                }
            }
        }
    }
}
```
This means that the code can iterate over the keys of `'default_github_project_fields'` looking for matches with the upstream fields; when it finds a match, it uses the corresponding value to determine the downstream field to populate by looking at the `'fieldmap'` value, and, if the `'options'` (or `'valuemap'`) key is present, it uses the corresponding table to convert the upstream value to the downstream value (if the key is not present, then the value is propagated unchanged).

Hopefully this structure is flexible enough that the code doesn't have to know anything about individual fields on either the upstream or downstream sides and that everything it needs is in these tables.

kpiwko:
This way we can query values and field names at the same time.

```
query GithubIssueDetails($orgname: String!, $reponame: String!, $issuenumber: Int!) {
  repository(owner: $orgname, name: $reponame) {
    issue(number: $issuenumber) {
      title
      body
      projectItems(first: 1) {
        nodes {
          fieldValues(first: 10) {
            nodes {
              ... on ProjectV2ItemFieldSingleSelectValue {
                name
                fieldName: field {
                  ... on ProjectV2FieldCommon {
                    name
                  }
                }
              }
              ... on ProjectV2ItemFieldTextValue {
                text
                fieldName: field {
                  ... on ProjectV2FieldCommon {
                    name
                  }
                }
              }
              ... on ProjectV2ItemFieldNumberValue {
                number
                fieldName: field {
                  ... on ProjectV2FieldCommon {
                    name
                  }
                }
              }
              ... on ProjectV2ItemFieldDateValue {
                date
                fieldName: field {
                  ... on ProjectV2FieldCommon {
                    name
                  }
                }
              }
              ... on ProjectV2ItemFieldUserValue {
                users(first: 10) {
                  nodes {
                    login
                  }
                }
                fieldName: field {
                  ... on ProjectV2FieldCommon {
                    name
                  }
                }
              }
              ... on ProjectV2ItemFieldIterationValue {
                title
                duration
                startDate
                fieldName: field {
                  ... on ProjectV2FieldCommon {
                    name
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

sync2jira/downstream_issue.py
```diff
@@ -960,12 +960,21 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
-        try:
-            existing.update({jirafieldname: str(getattr(issue, name))})
-        except JIRAError:
-            # Add a comment to indicate there was an issue
-            client.add_comment(existing, f"Error updating GitHub project field")
+        fieldvalue = getattr(issue, name)
+        if name == 'storypoints':
+            jirafieldname =  next(iter(values['fieldmap'].values()))
+            try:
+                existing.update({jirafieldname: fieldvalue})
```
baijum, line 967: Need to create a work around for empty values: https://jira.atlassian.com/browse/JSDSERVER-4364

baijum: Update seems to be working: https://community.developer.atlassian.com/t/how-to-set-request-field-to-null-during-creation-when-a-default-value-is-defined/37425/2

baijum:
Based on offline discussion, the generalized configuration structure is not adopted for this PR.

sync2jira/upstream_issue.py
```diff
@@ -252,26 +305,30 @@ def github_issues(upstream, config):
...
+        github_project_fields = default_github_project_fields | project_github_project_fields
+        variables = {"orgname": orgname, "reponame": reponame, "issuenumber": issuenumber}
+        for fieldname, values in github_project_fields.items():
+            response = requests.post(graphqlurl, headers=headers, json={"query": ghquery, "variables": variables})
+            data = response.json()
+            try:
+                nodes = data['data']['repository']['issue']['projectItems']['nodes'][0]['fieldValues']['nodes']
+                for item in nodes:
+                    if item.get('fieldName') and item.get('fieldName').get('name'):
+                        ghFieldName = item.get('fieldName').get('name')
```
webbnh, line 323: The first condition in line 322 should be `'fieldName' in item`, and the second condition you reuse on the next line (and you don't need the first `get()` call in it), so I suggest:
```suggestion
                    if 'fieldName' in item:
                        ghFieldName = item['fieldName'].get('name')
                        if ghFieldName:
```
or
```suggestion
                    ghFieldName = item.get('fieldName', {}).get('name')
                    if ghFieldName:
```

webbnh (changes requested):
Do we handle the situation where the upstream value for story points or priority gets unset?  (Or, can this not happen?)

Is this PR missing the changes to [`fedmsg.d/sync2jira.py`](https://github.com/release-engineering/Sync2Jira/blob/main/fedmsg.d/sync2jira.py)?

sync2jira/upstream_issue.py
```diff
@@ -252,26 +305,30 @@ def github_issues(upstream, config):
...
+                        if next(iter(github_project_fields['priority']['fieldmap'].keys())) == ghFieldName:
+                            issue['priority'] = item.get('name')
+                        if next(iter(github_project_fields['storypoints']['fieldmap'].keys())) == ghFieldName:
+                            issue['storypoints'] = int(item.get('number'))
+            except (TypeError, KeyError) as err:
+                log.debug("Error fetching %s!r from GitHub %s/%s#%s: %s",
+                          ghFieldName, orgname, reponame, issuenumber, err)
```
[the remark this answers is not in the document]
baijum: golang influence ;-)

sync2jira/upstream_issue.py
```diff
@@ -309,26 +311,26 @@ def github_issues(upstream, config):
         issue['priority'] = ''
         orgname, reponame = upstream.rsplit('/', 1)
         issuenumber = issue['number']
-        default_github_project_fields = config['sync2jira']['default_github_project_fields']
-        project_github_project_fields = config['sync2jira']['map']['github'][upstream]['github_project_fields']
-        github_project_fields = default_github_project_fields | project_github_project_fields
+        github_project_fields = config['sync2jira']['map']['github'][upstream]['github_project_fields']
```
webbnh, line 314: We're now requiring _every_ project to provide the field mappings?  There's no common practice that we can put into a default?

sync2jira/upstream_issue.py
```diff
@@ -309,26 +311,26 @@ def github_issues(upstream, config):
...
+        if response.status_code != 200:
+            log.debug("Error fetching from GitHub: %s", response.text)
+            continue
+        data = response.json()
+        try:
+            nodes = data['data']['repository']['issue']['projectItems']['nodes'][0]['fieldValues']['nodes']
+            for item in nodes:
+                if 'fieldName' in item:
+                    gh_field_name = item.get('fieldName', {}).get('name')
+                    if gh_field_name in github_project_fields['priority']['fieldmap']:
```
webbnh, line 326: Given the `if` statement at line 326 and the `get()` with the default value of `{}` in line 325, you don't need the `if` statement at line 324.

On the other hand, if you like and want to keep the if statement at line 324, then you should change line 325 to
```python
                    gh_field_name = item['fieldName'].get('name')
```
That is, if control reaches line 325, then we know that `'fieldName'` is a valid key in `item`, so we don't need to call `get()` to try to get its value -- we can index it unconditionally.

sync2jira/upstream_issue.py
```diff
@@ -309,26 +311,26 @@ def github_issues(upstream, config):
...
+                        issue['priority'] = item.get('name')
+                    if gh_field_name in github_project_fields['storypoints']['fieldmap']:
```
webbnh, line 328: Since (I think) you've gotten rid of the default field mappings, can we be sure that `'priority'` is a key in `github_project_fields` at line 326?  Don't we need a check for that?

Analogously for `'storypoints'` at line 328.

tests/test_intermediary.py
```diff
@@ -74,7 +74,6 @@ def checkResponseFields(self, response):
         self.assertEqual(response.upstream, 'github')
         self.assertEqual(response.comments, [{'body': 'mock_body', 'name': 'mock_name', 'author': 'mock_author',
                                               'changed': None, 'date_created': 'mock_date', 'id': 'mock_id'}])
-        self.assertEqual(response.priority, None)
```
webbnh, line 76: Should this assertion be replaced with something else?  (Is simply deleting it the correct thing to do?)

tests/test_upstream_issue.py
```diff
@@ -15,9 +15,6 @@ class TestUpstreamIssue(unittest.TestCase):
     def setUp(self):
         self.mock_config = {
             'sync2jira': {
-                'default_github_project_fields': {
-                    'storypoints': ('Estimate', 'customfield_12310243')
-                },
                 'map': {
                     'github': {
                         'org/repo': {'sync': ['issue'], 'github_project_fields': {}},
```
webbnh, line 20: You've removed the `default_github_project_fields` definition (and concept), but what remains here is an empty `github_project_fields`.  Shouldn't we have a test case which tests some non-empty ones??

tests/test_upstream_issue.py
```diff
@@ -26,7 +23,6 @@ def setUp(self):
                 'jira': {
                     # Nothing, really..
                 },
-                'default_github_project_fields': {},
```
webbnh: Ha!  This was superseding the definition at line 18....  😛

fedmsg.d/sync2jira.py
```diff
@@ -53,7 +53,6 @@
                 'token_auth': 'YOUR_JIRA_ACCESS_TOKEN',
             },
         },
-        'default_github_project_fields': {'storypoints': ('Estimate', 'customfield_12310243')},
         'map': {
             'github': {
                 'GITHUB_USERNAME/Demo_project': {'project': 'FACTORY', 'component': 'gitbz',
```
webbnh, line 58: Shouldn't there be a `'github_project_fields'` key in line 59, now?

sync2jira/upstream_issue.py
webbnh: I suspect that this is the source of the unit test failures.

webbnh (changes requested):
With the removal of the default mappings, now every repo needs to provide the mappings for the fields that it requests to be sync'd, and we have to be prepared for the mappings to be missing.  (And, we should test the various combinations:  no sync with no mapping specified; sync requested with mapping specified; sync requested with missing mapping; and no sync with unused mapping (just to prove that it works).

Do we handle the situation where the upstream value for story points or priority gets unset? (Or, can this not happen?)

sync2jira/downstream_issue.py
```diff
@@ -365,7 +366,7 @@ def _find_comment_in_jira(comment, j_comments):
             else:
                 # Else they are equal and we can return the item
                 return item
-        if comment['date_created'] < UPDATE_DATE:
+        if comment['date_created'].replace(tzinfo=timezone.utc) < UPDATE_DATE.replace(tzinfo=timezone.utc):
```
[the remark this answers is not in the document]
ralphbean: I'm taking a stab at this over at https://github.com/release-engineering/Sync2Jira/pull/222

tests/test_intermediary.py
baijum: This assertion was part of `checkResponseFields(...)` function which includes common assertions.  This worked until now as the value for `priority` was indeed `None`, because there some previous attempt to implement the priority field syncing. This PR introduce `priority` syncing, so common assertion will not work anymore.

sync2jira/upstream_issue.py
baijum: Fixed.

sync2jira/upstream_issue.py
```diff
@@ -346,6 +346,11 @@ def github_issues(upstream, config):
         yield issue
 
 
+def get_github_project_details(graphqlurl, headers, ghquery, variables):
+    response = requests.post(graphqlurl, headers=headers, json={"query": ghquery, "variables": variables})
+    return response
```
webbnh, line 351: I don't see any benefit to encapsulating the call to `post()` inside a helper function.  Why did you do this?

If it is just for mocking, it's as easy to mock `requests.post()` as it is to mock `get_github_project_details()`.  There is [a nice example here](https://github.com/release-engineering/Sync2Jira/blob/18dce8e31431fef4f484e2eab9cbc70f5088a2b6/tests/test_main.py#L510-L513) (albeit with `get()` instead of `post()`, but it works the same).

sync2jira/upstream_issue.py
```diff
@@ -257,26 +311,30 @@ def github_issues(upstream, config):
...
+        response = get_github_project_details(graphqlurl, headers, ghquery, variables)
+        if response.status_code != 200:
+            log.debug("Error fetching from GitHub: %s", response.text)
+            continue
+        data = response.json()
+        try:
+            nodes = data['data']['repository']['issue']['projectItems']['nodes'][0]['fieldValues']['nodes']
+            for item in nodes:
+                if 'fieldName' in item:
+                    gh_field_name = item.get('fieldName', {}).get('name')
```
webbnh, line 329: Given that, if execution reaches line 329, we know that `'fieldName'` is in `item`, thanks to the `if` statement at line 328, we don't need the first `get()` call:
```suggestion
                if 'fieldName' in item:
                    gh_field_name = item['fieldName'].get('name')
```

webbnh (commented):
A couple of comments.

sync2jira/upstream_issue.py
baijum: Fixed in 725a04868080b4dd7aebced4a40a6b28bb8725e5

sync2jira/upstream_issue.py
baijum: Fixed in 725a04868080b4dd7aebced4a40a6b28bb8725e5

fedmsg.d/sync2jira.py
```diff
@@ -53,7 +53,9 @@
                 'token_auth': 'YOUR_JIRA_ACCESS_TOKEN',
             },
         },
-        'default_github_project_fields': {'storypoints': ('Estimate', 'customfield_12310243')},
+        'default_jira_fields': {
+            'storypoints': 'customfield_12310243',
+            },
```
baijum, line 58: @webbnh @ralphbean @Zyzyx I am changing this configuration. This is not backward compatible. So, wherever sync2jira is running, the configuration needs to be changed. Existence of the old `default_github_project_fields` will not be a problem.

I made this change as there is no need to specify the GitHub field as it is not specific to a particular Jira instance. As these fields are unique to a particular Jira instance, it would be more appropriate to refer to them as `default_jira_fields`.

Here is a complete example:

```
"default_jira_fields": {
  "storypoints": "customfield_12310243"
},

"map": {
  "github": {
    "developerproductivity/open-connectors": {
      "project": "TEST",
      "component": "kernel",
      "sync": [
        "pullrequest",
        "issue"
      ],
      "issue_updates": [
        "comments",
        "github_project_fields"
      ],
      "github_project_fields": {
        "storypoints": {
          "gh_field": "Estimate"
        },
        "priority": {
          "gh_field": "Priority",
          "options": {
            "P0": "Blocker",
            "P1": "Critical",
            "P2": "Major",
            "P3": "Minor",
            "P4": "Optional",
            "P5": "Trivial"
          }
        }
      }
    }
  }
},
```

cc. @kpiwko @mattreid

tests/test_intermediary.py
```diff
@@ -94,6 +94,10 @@ def test_from_github_open(self):
         self.checkResponseFields(response)
 
         self.assertEqual(response.fixVersion, ['mock_milestone'])
+        self.assertEqual(response.priority, 'mock_priority')
+        self.assertEqual(response.content, 'mock_content')
+        self.assertEqual(response.reporter, 'mock_reporter')
+        self.assertEqual(response.assignee, 'mock_assignee')
```
webbnh, line 100: Can some or all of these assertions be added to `checkResponseFields()` instead of repeating them in each test?

tests/test_upstream_issue.py
```diff
@@ -15,12 +15,9 @@ class TestUpstreamIssue(unittest.TestCase):
...
         self.mock_config = {
             'sync2jira': {
-                'default_github_project_fields': {
-                    'storypoints': ('Estimate', 'customfield_12310243')
-                },
                 'map': {
                     'github': {
-                        'org/repo': {'sync': ['issue']},
+                        'org/repo': {'sync': ['issue'], 'github_project_fields': {}},
                     },
```
webbnh, line 21: So, by default, we're testing with an empty `'github_project_fields'`, shouldn't we have tests where it has entries in it?

tests/test_upstream_issue.py
```diff
@@ -97,12 +94,16 @@ def setUp(self):
         self.mock_github_client.get_repo.return_value = self.mock_github_repo
         self.mock_github_client.get_user.return_value = self.mock_github_person
 
+        # Mock GitHub project response
```
webbnh, line 97: This looks like an orphaned comment...should it be moved or deleted?  Or...is there content which is supposed to be added here?

sync2jira/downstream_issue.py
```diff
@@ -977,12 +980,27 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
+                client.add_comment(existing, f"Error updating GitHub project storypoints field: {err}")
+        elif name == 'priority':
+            jira_priority = github_project_fields['priority']['options'].get(fieldvalue)
+            if not jira_priority:
+                jira_priority = ''
+            try:
+                existing.update({'priority': {'name': jira_priority}})
+            except JIRAError as err:
+                # Add a comment to indicate there was an issue
+                client.add_comment(existing, f"Error updating GitHub project priority field: {err}")
```
webbnh, line 1003: Is there unit test coverage for this code?

sync2jira/upstream_issue.py
```diff
@@ -257,26 +311,37 @@ def github_issues(upstream, config):
         if issue.get('milestone', None):
             issue['milestone'] = issue['milestone']['title']
 
-        if not issue.get('storypoints', None):
-            issue['storypoints'] = ''
+        issue_updates = config['sync2jira']['map']['github'][upstream].get('issue_updates', {})
+        if 'github_project_fields' in issue_updates:
+            issue['storypoints'] = None
+            issue['priority'] = ''
```
webbnh, line 317: Is there unit test coverage for this code? (I.e., lines 314-344)

webbnh (changes requested):
Baiju, this is great stuff, and it's very nearly ready to merge, but I have a bunch of comments.  In particular, there are some things that need to be addressed before this is merged:

- [x] Foremost, there is a `breakpoint()` call which probably should have been removed when you finished debugging; I don't think we want that merged into `main`.
- [x] The `fedmsg.d/sync2jira.py` contents seem to have grown stale, and this PR continues the drift.
- [x] Regarding your [comment](https://github.com/release-engineering/Sync2Jira/pull/216#discussion_r1816770782), I suspect that we should avoid breaking backward compatibility, and I think it is straightforward to do so.
- [ ] I called out a couple of places where we should probably be more paranoid, to try to prevent _Sync2Jira_ from crashing due to external problems.
- [ ] I'm concerned about the unit test coverage -- I suspect that a lot of the new code isn't actually tested by the unit tests.

There are some lower priority things which you should consider improving for this PR:
- [ ] There is a log message which should probably include more context.
- [ ] The patching for `requests.post` in the tests should probably be done slightly differently to match existing practices.
- [ ] And, there are a bunch of smaller suggestions and nits.


Also, I've raised a concern about the source of Priority and Story Point values not being unique.  There is probably nothing that we can do about this, so it should presumably be called out in the documentation.

fedmsg.d/sync2jira.py
baijum: This is an unrelated issue. I think somebody has forgotten to change the `updates` when they changed the structure. And `github_project_fields` is optional, even if it is missing, that doesn't matter, because the rest of the code doesn't show a full example.  I can create an issue (and separate PR) to update this part. Let me know if that's okay.

webbnh: > I think somebody has forgotten to change the updates when they changed the structure.

Agreed.

> And `github_project_fields` is optional, even if it is missing, that doesn't matter, because the rest of the code doesn't show a full example.

But, that's part of my concern:  this _is_ as far as I can tell, _example_ code, and so completely omitting a feature like `github_project_fields` from it doesn't serve the customer well.

And, given that you're already making changes to the module, why not bring it up to date, given that it is trivial to do so?  (You've already pasted an example configuration into [a nearby comment](https://github.com/release-engineering/Sync2Jira/pull/216/files?w=0#r1816770782), why not just paste it into the module?)

> I can create an issue (and separate PR) to update this part. Let me know if that's okay.

I wouldn't block this PR solely on this issue; it just seems like creating an issue and a separate PR would be more effort (given that you're presumably already going to update this PR) than just making the requested change here.

baijum: The current `issues` only has an Ellipsis (...) as a dummy value to indicate that there many possible values.
I can change it to `'issue_updates': [...]` or do you want me to include ass possible values (this may increase in future). After some searching, I found these many available values:
```
issue_updates=[
  'comments',
  'upstream_id',
  'title',
  'description',
  'github_markdown',
  'upstream_id',
  'url',
  {'transition': 'Closed'},
  {'assignee': {'overwrite': False}},
  'github_project_fields'
],
```
Please suggest.

baijum: For now, I have changed it to `issue_updates` with a list containing an Ellipsis. Let me know, if all possible values needs to be included.

sync2jira/upstream_issue.py
```diff
@@ -311,30 +311,37 @@ def github_issues(upstream, config):
...
+                        gh_field_name = item['fieldName'].get('name')
+                        if 'priority' in github_project_fields and gh_field_name in github_project_fields['priority']['gh_field']:
+                            issue['priority'] = item.get('name')
+                        if 'storypoints' in github_project_fields and gh_field_name in github_project_fields['storypoints']['gh_field']:
+                            issue['storypoints'] = int(item.get('number'))
+            except (TypeError, KeyError) as err:
+                breakpoint()
+                log.debug("Error fetching %s!r from GitHub %s/%s#%s: %s",
+                          gh_field_name, orgname, reponame, issuenumber, err)
+                continue
```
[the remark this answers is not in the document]
baijum: Fixed. cd3ac60e20347699dd8307613ec4e5090d9b4ff6

sync2jira/downstream_issue.py
```diff
@@ -703,7 +704,7 @@ def _create_jira_issue(client, issue, config):
 
     # Update relevant information (i.e. tags, assignees etc.) if the
     # User opted in
-    _update_jira_issue(downstream, issue, client)
+    _update_jira_issue(downstream, issue, client, default_jira_fields)
```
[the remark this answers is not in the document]
baijum: Fixed. cd3ac60e20347699dd8307613ec4e5090d9b4ff6

sync2jira/upstream_issue.py
```diff
@@ -311,30 +311,37 @@ def github_issues(upstream, config):
...
+            github_project_fields = config['sync2jira']['map']['github'][upstream]['github_project_fields']
+            variables = {"orgname": orgname, "reponame": reponame, "issuenumber": issuenumber}
+            response = requests.post(graphqlurl, headers=headers, json={"query": ghquery, "variables": variables})
+            if response.status_code != 200:
+                log.debug("Error fetching from GitHub: %s", response.text)
+                continue
+            data = response.json()
+            gh_issue = data['data']['repository']['issue']
+            if not gh_issue:
+                log.debug("Error fetching issue from GitHub: %s", response.text)
```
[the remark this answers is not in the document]
baijum: This is the current text:
```
DEBUG: Error fetching issue from GitHub: {"data":{"repository":{"issue":null}},"errors":
[{"type":"NOT_FOUND","path":["repository","issue"],"locations":[{"line":6,"column":11}],
"message":"Could not resolve to an Issue with the number of 4."}]}
```
I this is better than `data['errors']` which misses that easy to understand message and repository details.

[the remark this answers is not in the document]
baijum: The `data` key will be always present, if the status code is `200` (see my previous comment for a typical error).
Also, it would be possible to access `issue` like this: `data['data']['repository']['issue']`

So I would prefer to keep this code like this.

tests/test_intermediary.py
baijum: Done. ce912603abb8262570d7ff20c113da839bfac19a

sync2jira/downstream_issue.py
baijum: Done. 7db5135c032ea8c8a5684d9b108863acc9a5e5a5

tests/test_upstream_issue.py
```diff
@@ -94,10 +94,18 @@ def setUp(self):
...
-        # Mock GitHub project response
+    def tearDown(self):
+        self.mock_config = {}
+        self.mock_github_comment = MagicMock()
+        self.mock_github_message = {}
+        self.mock_github_issue = MagicMock()
+        self.mock_github_issue_raw = {}
+        self.mock_github_person = MagicMock()
+        self.mock_github_repo = MagicMock()
+        self.mock_github_client = MagicMock()
```
webbnh, line 105: If this is supposed to be resetting these values back to what they were given in `setUp()`, consider having `setUp()` call this function instead of relying on separate assignments.  (That way, the two cannot drift apart.)

webbnh (commented):
Hi Baiju, here are some comments.  I didn't quite finish, but I figured I should give you what I've got.  I'll try to take this up again tonight or tomorrow.

tests/test_upstream_issue.py
baijum: `tearDown` is called by the unitest framework. We don't need to call it explicitly. The execution order is :
```
setup()
test_someting()
tearDown()
```

tests/test_intermediary.py
```diff
@@ -95,9 +95,6 @@ def test_from_github_open(self):
 
         self.assertEqual(response.fixVersion, ['mock_milestone'])
         self.assertEqual(response.priority, 'mock_priority')
-        self.assertEqual(response.content, 'mock_content')
-        self.assertEqual(response.reporter, 'mock_reporter')
-        self.assertEqual(response.assignee, 'mock_assignee')
```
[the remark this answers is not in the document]
baijum: These assertions are part of `checkResponseFields()` and it's called in line no. 98.

tests/test_downstream_issue.py
```diff
@@ -1672,3 +1673,35 @@ def test_update_on_close_no_config(self):
...
         self.mock_downstream.update.assert_not_called()
+
+    @mock.patch('jira.client.JIRA')
+    def test_update_github_project_fields_storypoints(self, mock_client):
+        """
+        """
+        mock_github_project_fields = {
+         "storypoints": {
+           "gh_field": "Estimate"
+         }}
```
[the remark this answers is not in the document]
baijum: Fixed. f228986740274362797393dc25c6c7f67ee17f69

tests/test_downstream_issue.py
```diff
@@ -1672,3 +1673,35 @@ def test_update_on_close_no_config(self):
 
         # Assert everything was called correctly
         self.mock_downstream.update.assert_not_called()
+
+    @mock.patch('jira.client.JIRA')
+    def test_update_github_project_fields_storypoints(self, mock_client):
+        """
+        """
```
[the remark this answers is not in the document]
baijum: Fixed. f228986740274362797393dc25c6c7f67ee17f69

sync2jira/upstream_issue.py
```diff
@@ -328,17 +333,38 @@ def github_issues(upstream, config):
...
+                            if 'fieldName' in item:
+                                gh_field_name = item['fieldName'].get('name')
+                                if 'priority' in github_project_fields\
+                                        and gh_field_name in github_project_fields['priority']['gh_field']:
+                                    issue['priority'] = item.get('name')
+                                if 'storypoints' in github_project_fields\
+                                        and gh_field_name in github_project_fields['storypoints']['gh_field']:
+                                    issue['storypoints'] = int(item.get('number'))
+                if not found:
+                    log.debug("Issue is not associated with the configured project, but other association exists.")
```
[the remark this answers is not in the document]
baijum: Done. bac05356daac0ebb4b80a825fc92307d1453e6a9

sync2jira/upstream_issue.py
```diff
@@ -328,17 +333,38 @@ def github_issues(upstream, config):
...
+                    log.debug("Issue %s is not associated with any project", issuenumber)
+                    continue
+                if not configured_project_id and len(projectItemNodes) > 1:
+                    project_ids = ", ".join([":".join([x['project']['id'], x['project']['title']]) for x in projectItemNodes])
+                    log.debug("Project ID is not configured.")
+                    log.debug("And the issue %s (repo: %s) is associated with more than one projects: %s",
+                              issuenumber, upstream, project_ids)
+                    continue
+                if not configured_project_id and len(projectItemNodes) == 1:
+                    configured_project_id = projectItemNodes[0]['project']['id']
```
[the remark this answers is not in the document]
baijum: Done. 7c53fbbc07afc4176040f0d77e7a9214dce783f7

sync2jira/upstream_issue.py
```diff
@@ -328,17 +333,38 @@ def github_issues(upstream, config):
...
-                            issue['storypoints'] = int(item.get('number'))
+                projectItemNodes = gh_issue['projectItems']['nodes']
+                if len(projectItemNodes) == 0:
+                    log.debug("Issue %s is not associated with any project", issuenumber)
+                    continue
+                if not configured_project_id and len(projectItemNodes) > 1:
+                    project_ids = ", ".join([":".join([x['project']['id'], x['project']['title']]) for x in projectItemNodes])
+                    log.debug("Project ID is not configured.")
+                    log.debug("And the issue %s (repo: %s) is associated with more than one projects: %s",
+                              issuenumber, upstream, project_ids)
```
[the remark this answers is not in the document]
baijum: This is a debug log for contributors of this project. Normally this is turned of during production. And this is only useful during development. We can assume that developers know how to access ID.

fedmsg.d/sync2jira.py
baijum: Done. fd926afa5a5c00d29d0062717929887da89a6f6f

tests/test_upstream_issue.py
baijum: removed

sync2jira/upstream_issue.py
baijum: added test cases.

tests/test_upstream_issue.py
baijum: I have added tests for both the story points and priority.

sync2jira/upstream_issue.py
baijum: The logic has been changed in such a way that first check for the flag 'github_project_fields' inside 'issue_updates'. If it is present, the value for `github_project_fields` is expected.

sync2jira/upstream_issue.py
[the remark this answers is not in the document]
webbnh: I concur that, for the developers of _Sync2Jira_, the Project ID might be more convenient to work with, and so I can see how/why you might want to print that in the code cited here.

However, as I understand it, the config file will need to include a reference to the GitHub project in order to find the appropriate issue priority and story points (see [`upstream_issue.py#L319`](https://github.com/release-engineering/Sync2Jira/blob/ce419ad9fe7b74433d15854ae37a50358d883399/sync2jira/upstream_issue.py#L319)).  Whom do you expect to add that to the config file, and how do you expect them to obtain the ID?

_That's_ the value that I think should be the GH Project URL.  (What you print here is of secondary concern.)

tests/test_upstream_issue.py
webbnh: Right, but if something is changed in `setUp()`, how can we be sure that the appropriate change is made to `tearDown()` to match it?

That is, rather than having identical assignments like
```python
        self.mock_github_repo = MagicMock()
```
made in _both_ `setUp()` **and** `tearDown()`, why not have `setUp()` explicitly call `tearDown()` and then remove the duplicated calls from `setUp()`?

That way, we won't have duplicated code, and it will be more obvious to anyone reading (i.e., changing) `setUp()` that there is a direct relationship with the contents of `tearDown()`.

docs/source/config-file.rst
```diff
@@ -61,6 +61,15 @@ The config file is made up of multiple parts
...
   Ensure to name them appropriately, in name of the JIRA instance above is `example`.
 
+.. code-block:: python
+
+    "default_jira_fields": {
+      "storypoints": "customfield_12310243"
+    },
+
+* Specify the jira field for syncing. As of now the supported field is `storypoints`. It is also possible to
+  add `priority`, but in the abscence the default value will be `priority`.
```
webbnh, line 71: "abscence" ==> "absence"

I might insert "only" before "supported", but I think the third sentence could be clearer.  I would be inclined to rework this whole description.  I think we ought to describe this mapping in terms of `github_project_fields`.  That is, the keys in `default_jira_fields` correspond to the values available to be specified in `github_project_fields`, and the keys' values provide the names of the upstream fields which are used to fetch the values for those operations.

Currently, we support only `"storypoints"` and `"priority"`.  By default, the value for the priority is sourced from the upstream `"priority"` field, but providing a `"priority"` key for `default_jira_fields` will override that.  Unfortunately, there is no default available for sourcing story points, so, the `"storypoints"` key must be provided if projects wish to sync story point values, and, for our corporate Jira instance, the value should be specified as `"customfield_12310243"`.

webbnh (commented):
Sorry, Baiju...I've run out of time for today; I'll try again tomorrow.

tests/test_upstream_issue.py
baijum: Removed `tearDown` 569871c6a9a6a01e20dc224dedb67ed13c5c885f.
Initialization is already happening in `setUp`.

docs/source/config-file.rst
baijum: Done. e4272df08a24e90b08afac9e97bb198e7baa37bc

sync2jira/upstream_issue.py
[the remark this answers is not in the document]
baijum: Done. 0105845e748dd6cc9197700b9cc2d603b34c6277

sync2jira/upstream_issue.py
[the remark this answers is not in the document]
webbnh: > The `data` key will be always present, if the status code is `200`

A tiny amount of testing seems to confirm this.

> it would be possible to access `issue` like this: `data['data']['repository']['issue']`

Actually, if the repository is not found by the query, the `data['data']['repository']` will be `null` or `None`, in which case referencing `data['data']['repository']['issue']` will result in a `TypeError`.  However, since, for the moment, Sync2Jira runs as the result of a message from upstream, I think we can assume that the repository will be found.  So, we only need to worry about whether the issue is found (and, again, if we're being driven by a message from upstream, it would be weird if it weren't).

> I would prefer to keep this code like this.

That's fine.

webbnh (commented):
I've got suggestions for another pass at the documentation and example and another pass at the project ID question.  (And a few other things.)

sync2jira/upstream_issue.py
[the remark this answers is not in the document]
baijum: We are checking if the value of issue is None. Since we are a using this GraphQL, I think the key `issue` would be always present:
```
...
repository(owner: $orgname, name: $reponame) {
  issue(number: $issuenumber) {
...
```

baijum:
> So, if a project sets config['sync2jira']['map']['github'][upstream]['issue_updates'] to include 'github_project_fields', then that project also has to provide its own copy of config['sync2jira']['map']['github'][upstream]['github_project_fields'], even if it's the same as all the other projects?

Yes. Since we are starting with the repository as the key, this will need to be repeated. The only thing that is not repeated is the Jira field mapping.

sync2jira/upstream_issue.py
```diff
@@ -260,6 +260,18 @@ def github_issues(upstream, config):
...
+        project_number = int(project_number_text)
+    except ValueError:
+        log.debug("Error input for project number: '%s'", project_number_text)
+        return
+    orgname, reponame = upstream.rsplit('/', 1)
+    try:
+        configured_project_id = _get_project_id(project_number, headers, orgname)
+    except Exception:
+        log.debug("Cannot access project ID")
+        return
```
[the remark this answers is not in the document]
baijum: I have changed to use the project number instead of the project ID.

sync2jira/downstream_issue.py
```diff
@@ -976,13 +978,33 @@ def _update_github_project_fields(client, existing, issue, github_project_fields
...
+                # Add a comment to indicate there was an issue
+                client.add_comment(existing, f"Error updating GitHub project storypoints field: {err}")
+        elif name == 'priority':
+            try:
+                jirafieldname = default_jira_fields['priority']
+            except KeyError:
+                jirafieldname = 'priority'
+            jira_priority = github_project_fields['priority']['options'].get(fieldvalue)
+            if not jira_priority:
+                jira_priority = ''
```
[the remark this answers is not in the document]
baijum: I have taken the previous suggestion.

sync2jira/upstream_issue.py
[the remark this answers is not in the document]
webbnh: > I think the key `issue` would be always present:

Only if there is a hit on `repository` -- if the repository is not found, then _it's_ value will be `None` and there won't be anywhere for the `issue` key to be stored in the response.  But, as I said, since this code is triggered by a message which originates from the repository, I don't think we need to worry about the repository disappearing before we can issue the query.  🙂

> We are checking if the value of issue is `None`.

Not to be pedantic, but sometimes the difference matters:  we checking if the value of `gh_issue` is "false-y".  It is correct that `None` is a false-y value, but so is `{}` (along with `[]` and `''`), and there could be a big difference between `None` and `{}`.  🥴

webbnh (commented):
I've resolved my comments from my earlier reviews (or, hidden them, technically) -- we're really close -- but your latest update prompted some comments, unfortunately.

The foremost is, I think we've got an undesirable behavior when the user doesn't have a project number configured and there are multiple issues which are associated with various projects.

The other comments center around the handling of the project ID.  Using the number is much better.  However, I'm concerned that the number by itself isn't sufficiently unique.  As an alternative, the project URL is readily available both to the user and to the tool via the query, and it contains both the project owner and the number which makes it properly unique.  (And, it doesn't require conversion, extraction, or construction in order to compare.)  (However, if you're not interested in using the URL, I've suggested that you compare the numbers as strings, because it avoids us having to deal with conversion errors.)

Other than those, there are just some nits around some log messages.

webbnh (changes requested):
I think there are a few things which need to be addressed before this is merged:

- [x] There is an explicit `breakpoint()` call which should presumably be removed.
- [ ] The caller of `_get_current_project_node()` needs to be prepared for a return value of `None`.

I wouldn't object to the merge on this basis, but I really don't think that unit tests should be relying on logging calls.

Beyond those, I have some coding suggestions.

sync2jira/downstream_issue.py
```diff
@@ -997,12 +997,14 @@ def _update_github_project_fields(client, existing, issue,
...
             except KeyError:
                 jirafieldname = 'priority'
-            jira_priority = values.get('options', {}).get(fieldvalue, '')
-            try:
-                existing.update({jirafieldname: {'name': jira_priority}})
-            except JIRAError as err:
-                # Add a comment to indicate there was an issue
-                client.add_comment(existing, f"Error updating GitHub project priority field: {err}")
+            jira_priority = values.get('options', {}).get(fieldvalue)
+            if jira_priority:
```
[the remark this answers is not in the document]
baijum: I noticed that setting empty string is raising an error. The priority must have some value.

sync2jira/upstream_issue.py
```diff
@@ -260,12 +261,7 @@ def github_issues(upstream, config):
     # And get comments if needed
     github_client = Github(config['sync2jira']['github_token'], retry=5)
 
-    project_number_text = config['sync2jira']['map']['github'][upstream].get('github_project_number')
-    try:
-        project_number = int(project_number_text)
-    except ValueError:
-        log.debug("Error input for project number: '%s'", project_number_text)
-        project_number = 0
+    project_number = config['sync2jira']['map']['github'][upstream].get('github_project_number')
```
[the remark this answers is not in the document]
baijum: That will be inside the loop. So, I have moved just before the loop.

tests/test_upstream_issue.py
```diff
@@ -615,6 +615,54 @@ def test_fetch_github_data(self, mock_requests):
...
+        result = u._get_current_project_node('org/repo', None, 'mock_number', gh_issue)
+        self.assertEqual(result, {'project': {'number': 1}})
+        mock_log.debug.assert_not_called()
+
+    @mock.patch(PATH + 'log')
+    def test_get_current_project_node_multiple_projects_no_project_number(self, mock_log):
+        """
+        This function tests '_get_current_project_node' where there are multiple project nodes and no project number
+        """
+        gh_issue = {'projectItems': {'nodes': [{'project': {'number': 1, 'url': 'url1', 'title': 'title1'}}, {'project': {'number': 2, 'url': 'url2', 'title': 'title2'}}]}}
```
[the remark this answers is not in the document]
baijum: Done

tests/test_upstream_issue.py
```diff
@@ -615,6 +615,54 @@ def test_fetch_github_data(self, mock_requests):
...
+
+    @mock.patch(PATH + 'log')
+    def test_get_current_project_node_single_project_no_project_number(self, mock_log):
+        """
+        This function tests '_get_current_project_node' where there is a single project node and no project number
+        """
+        gh_issue = {'projectItems': {'nodes': [{'project': {'number': 1}}]}}
+        result = u._get_current_project_node('org/repo', None, 'mock_number', gh_issue)
+        self.assertEqual(result, {'project': {'number': 1}})
+        mock_log.debug.assert_not_called()
```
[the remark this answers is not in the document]
baijum: removed log assertions.

sync2jira/upstream_issue.py
```diff
@@ -386,6 +364,34 @@ def github_issues(upstream, config):
...
+                      issuenumber, upstream, prj)
+            return
+        for projectNode in projectItemNodes:
+            if projectNode['project']['number'] == project_number:
+                currentProjectNode = projectNode
+                break
+        else:
+            log.debug("Issue is not associated with the configured project, but other associations exists.")
+            return
+    return currentProjectNode
```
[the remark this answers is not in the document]
baijum: adopted this change with some minor changes.

sync2jira/upstream_issue.py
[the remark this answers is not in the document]
webbnh: That's a big loop!  😢  I'm afraid that this code is ripe for refactoring....

Thanks for trying....  🥴

webbnh (approved):
It looks good to me!

(Thanks for your patience, Baiju!!)

ralphbean:
Trying to address the rpm signature scanning failure at https://github.com/release-engineering/Sync2Jira/pull/231

sync2jira/intermediary.py
```diff
@@ -103,7 +102,7 @@ def from_github(cls, upstream, issue, config):
             comments=comments,
             tags=issue['labels'],
             fixVersion=[issue['milestone']],
-            priority=None,
+            priority=issue['priority'],
```
webbnh, line 105: So, I guess this should be,
```suggestion
            priority=issue.get('priority'),
```
eh?  😊

Issue #214, opened 2024-10-03
Filter: consider upstream GitHub issues matching one of a set of labels  Allow a list of labels 


yrodiere:
The `filter` feature allows limiting the sync to upstream issues that have a given "tag" (label): https://github.com/release-engineering/Sync2Jira/blob/090725f5bf6a6338ec0b74d6d4e4d94d3e817c9f/docs/source/config-file.rst?plain=1#L163-L168

For Pagure, one can specify a list of "tags" (a set, actually), and any issue matching at least one tag will be synced: https://github.com/release-engineering/Sync2Jira/blob/090725f5bf6a6338ec0b74d6d4e4d94d3e817c9f/sync2jira/upstream_issue.py#L185-L197

For GitHub, however, sync2jira only supports a single label: https://github.com/release-engineering/Sync2Jira/blob/090725f5bf6a6338ec0b74d6d4e4d94d3e817c9f/sync2jira/upstream_issue.py#L71-L75

Let's align the GitHub implementation on the Pagure one?

mbladel:
Since https://github.com/release-engineering/Sync2Jira/pull/217 was merged, I believe we can now consider this fixed and close the issue.

yrodiere:
Indeed, thank you!

PR #217, opened 2024-10-04
#214 Fix filtering issues based on labels

mbladel:
This bug should address https://github.com/release-engineering/Sync2Jira/issues/214 when listening for new GH issues.

Also, it should fix applying the `labels` filter to the url for initialization that was not working. Before, we were appending the list to the query, e.g.:

`https://api.github.com/repos/quarkusio/quarkus/issues?labels=['area/hibernate-orm', 'area/hibernate-search']`

While the required format is a comma-separated list of labels, like:

`https://api.github.com/repos/quarkusio/quarkus/issues?labels=area/hibernate-orm,area/hibernate-search`


sync2jira/upstream_issue.py
```diff
@@ -85,8 +85,8 @@ def handle_github_message(msg, config, pr_filter=True):
         if key == 'labels':
             # special handling for label: we look for it in the list of msg labels
             actual = [label['name'] for label in msg['msg']['issue']['labels']]
-            if expected not in actual:
-                log.debug("Label %s not set on issue: %s", expected, upstream)
+            intersection = set(actual) & set(expected)
+            if not intersection:
                 return None
```
webbnh, line 90: Rather than instantiating the `list` so that it can be used to construct a `set`, why not just instantiate `actual` _as_ a `set`?  And, rather than asking "not intersection" how about asking "if disjoint"?
```python
            actual = {label['name'] for label in msg['msg']['issue']['labels']}
            if actual.isdisjoint(expected):
                return None
```
(Note that `isdisjoint()` accepts an iterable, so there is no need to instantiate a `set` from `expected`.)

sync2jira/upstream_issue.py
```diff
@@ -187,6 +187,13 @@ def github_issues(upstream, config):
         .get('github', {})\
         .get(upstream, {})
 
+    for key, expected in _filter.items():
+        if key == 'labels':
+            # We have to flatten the labels list to a comma-separated string
+            _filter[key] = ','.join(expected)
+        else:
+            _filter[key] = expected
```
webbnh, line 195: If we hit the `else` clause, the value of `_filter[key]` already _is_ `expected`, so there is no need to reassign it.  You should just omit the `else` clause.

Also, I don't think the `key` `'labels'` can appear in `_filter.items()` more than once, so there is no need to iterate over the dictionary, just look up the key and replace it:
```suggestion
    _filter['labels'] = ','.join(expected)
```

webbnh (commented):
Here are a couple of code-quality suggestions.

Also, you need to rebase your branch.

sync2jira/upstream_issue.py
mbladel: Sure, I had just copied the old pagure implementation from the issue, but of course this is much better

sync2jira/upstream_issue.py
mbladel: Right, applied your suggestion (with a check since the `'labels'` filter is optinal)

mbladel:
Thank you for the tips @webbnh, sorry for the rookie mistakes but my python is very rusty.

I also aligned the `upstream_pr` synchronization filter for labels to flatten the list, and I've adjusted unit tests to always pass a list for `'labels'` filter key. We could be lenient and check `isinstance(_filter['labels'],str)` and keep a single value, but since the documentation specifies `labels` should be a list I don't know if that's necessary.

 Let me know if there's anything else I can do.

sync2jira/upstream_issue.py
```diff
@@ -188,6 +188,13 @@ def github_issues(upstream, config):
         .get('github', {})\
         .get(upstream, {})
 
+    if 'labels' in _filter:
+        # We have to flatten the labels list to a comma-separated string
+        expected = _filter['labels']
+        if not isinstance(expected,str):
+            assert isinstance(expected,list)
```
webbnh, line 195: I think we are trying to work toward a more robust mode of operation for `sync2jira`, so having it die suddenly from an assertion failure would be undesirable.  I don't know enough to suggest what _should_ happen here (@ralphbean? @Zyzyx?), but, if the value is neither a `str` nor a list of `str`'s, perhaps it should be replaced with an empty string.

Also, I have to ask, why is this level of paranoia warranted for issues and not for PRs?  (The code you added  `sync2jira/upstream_pr.py` line 147 doesn't do the type checks.)

webbnh (approved):
> Thank you for the tips @webbnh, sorry for the rookie mistakes but my python is very rusty.

You're most welcome, and no worries about the "mistakes" -- the code was fine, I just offered some options for improvement.  And, that's what review is for, after all -- we all get to learn from the experience.

> I also aligned the upstream_pr synchronization filter for labels to flatten the list, and I've adjusted unit tests to always pass a list for 'labels' filter key.

Good catches.

> We could be lenient and check isinstance(_filter['labels'],str) and keep a single value, but since the documentation specifies labels should be a list I don't know if that's necessary.

I don't have an opinion on that.  (Well, I'm sure I could scare one up, like "being more lenient obliges us to do more testing, so we should be sure that the leniency delivers sufficient value to justify its cost." 😉)

The changes look good to me, but I do have some questions:

- Is the level of paranoia embodied by the type checks warranted?  And, if so, why is it implemented only for issues and not for PRs?
- If the type checks fail, are we sure we want to raise an exception rather than taking some sort of remedial action...since killing the tool will require restarting the service.

Also, on the topic of "rookie mistakes", once we've started reviewing a PR, it's best not to squash your branch -- that erases the history and it means that subsequent reviews can't focus just on the changes that you've made since the last pass.  (As a general rule of thumb, if you have to force-push your branch, you've probably done something wrong (with the obvious exception being when you rebase to a new tip of `main`).)

sync2jira/upstream_issue.py
mbladel: Ah sorry, I actually didn't mean to push this - I was just testing this locally :sweat_smile:

mbladel:
> Also, on the topic of "rookie mistakes", once we've started reviewing a PR, it's best not to squash your branch -- that erases the history and it means that subsequent reviews can't focus just on the changes that you've made since the last pass. (As a general rule of thumb, if you have to force-push your branch, you've probably done something wrong (with the obvious exception being when you rebase to a new tip of `main`).)

Ah sorry, I'm used to squashing chances to a single commit as to not push "review commits" as part of the main repository history - that's just how we operate in most of the repositories I contribute to as we do not use merge-commits and only rebase. I'll stop that from now.

I'm away for a couple of days, as soon as I'm back on my laptop I'll remove the type check.

webbnh:
> Ah sorry, I'm used to squashing changes to a single commit as to not push "review commits" as part of the main repository history

No worries.  I concur with your concern, but there are opportunities to squash the branch after the review is complete.  (In this case, it was not a problem because the change is nice and small; but in a large change, it can be kind of soul-crushing....)

> we do not use merge-commits and only rebase.

My preference, too, is a linear history for the `main` branch.  But, that's not strictly necessary for a development branch, particularly if the intention is to squash it before "merging" it (i.e., via a "fast-forward" merge, which amounts to a rebase and relabel).

> I'm away for a couple of days, as soon as I'm back on my laptop I'll remove the type check.

Thanks.  There's no deadline.  🙂

sync2jira/upstream_issue.py
mbladel: Removed this, now I always assume we have a list if the `labels` key exists.

webbnh (approved):

mbladel:
Thanks for the approval @webbnh, I've rebased the branch on `main` once again to resolve conflicts. Can we go ahead and merge this now?

webbnh:
@ralphbean, @Zyzyx?

PR #219, opened 2024-10-04
Fix unit test failures

webbnh:
Unfortunately, while the CI currently runs the unit tests via GitHub Actions, the results aren't reported to PRs, and failures don't block merges.  As such, a recent merge broke the unit tests, and subsequent merges have introduced additional failures.

This PR comprises remediations for the existing unit tests and brings `main` back into passing status.

The branch consists of a number small, topical commits for your reviewing pleasure.  Please pay especial attention to the last commit which contains a series of changes reversing specific assertions.  Looking at the code, I don't understand how these tests ever passed...so, if someone could please explain that and/or point out why my change should not be included, I would greatly appreciate it.

This change also includes a modest refactoring of the unit tests for the intermediary module which DRYs out the code a bit.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -950,6 +950,7 @@ def _update_jira_labels(issue, labels):
     issue.update(data)
     log.info('Updated %s tag(s)' % len(_labels))
 
+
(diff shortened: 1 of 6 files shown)


Zyzyx:
I'm helping!

PR #222, opened 2024-10-15
Use tz-aware datetimes consistently

ralphbean:


sync2jira/downstream_issue.py
```diff
@@ -37,7 +38,8 @@
 
 # The date the service was upgraded
 # This is used to ensure legacy comments are not touched
-UPDATE_DATE = datetime(2019, 7, 9, 18, 18, 36, 480291)
+UTC = ZoneInfo(key='UTC')
+UPDATE_DATE = datetime(2019, 7, 9, 18, 18, 36, 480291, UTC)
```
webbnh, line 42: Just FYI, although `datetime` doesn't provide time zone objects in general, [it does provide `UTC`](https://docs.python.org/3/library/datetime.html#datetime.UTC), which saves you from having to import from `zoneinfo` and rolling your own.

PR #229, opened 2024-11-05
Avoid mutating `_filter` dict for subsequent calls

mbladel:
Fixes the code introduced with https://github.com/release-engineering/Sync2Jira/pull/217 that flattens the labels list to a comma-separated string by avoiding mutating the original `_filter` object.


sync2jira/upstream_issue.py
```diff
@@ -188,13 +188,14 @@ def github_issues(upstream, config):
...
-        _filter['labels'] = ','.join(_filter['labels'])
-
     url = 'https://api.github.com/repos/%s/issues' % upstream
     if _filter:
-        url += '?' + urlencode(_filter)
+        # We have to flatten the labels list to a comma-separated string
+        url_filter = {
+            key: ','.join(expected) if key == 'labels' else expected
+            for key, expected in _filter.items()
+        }
```
webbnh, line 197: I don't think this is enough to "save" us.  The problem is that `join()` takes an _iterable_, and you're hoping/expecting that `expected` is a `list`; however, `str`'s are _also_ iterable, and so if `expected` is a(n already flattened) string, then `join()` won't object -- it'll just insert a comma between each character:
```
>>> ','.join("string")
's,t,r,i,n,g'
```

I think the question that we want to be asking is, is `_filter['labels']` a `list` (or maybe a `tuple`, but, definitely, is it _not_ a `str`)?  If so, then flatten it; otherwise, take it as it is.

That said, the fact that `_filter` could refer to part of `config` does suggest that we should be making a copy instead of mutating `_filter` itself (although, if the conditionalization were correct, mutating `_filter` might save us time on subsequent issues...).

I might be inclined to try code like this:
```suggestion
       if isinstance(_filter.get('labels'), list):
            _filter['labels'] = ','.join(_filter['labels'])
```
(Or, you can do the equivalent on a `deepcopy()` of `_filter`, if you're not comfortable mutating `config`.)

webbnh (commented):
While avoiding mutating the configuration is helpful, using a better conditional would probably produce a better result (one which would allow us to keep the mutated configuration, if we like).

Also, it would be good if you could add a unit test which detects this problem and _then_ fix it.  🙂

sync2jira/upstream_issue.py
mbladel: > think the question that we want to be asking is, is _filter['labels'] a list (or maybe a tuple, but, definitely, is it not a str)? If so, then flatten it; otherwise, take it as it is.

I had proposed checking if the type of `_filter['labels']` was string, and just taking as-is - but it was decided to just assume it's always a list. Since the check is easy and would also avoid repeated processing I agree we should do that.

> I might be inclined to try code like this:

That's fine as well, I wanted to be extra safe thus the new object - but this should work as well. I've updated the change with your suggestion, thanks.

webbnh: > it was decided to just assume it's always a list.

Clearly that was a mistake!  😀  (We all failed to notice that we were mutating the configuration object, such that we could end up receiving a string quite easily!)

> I wanted to be extra safe thus the new object

That's generally a good and appropriate thing to do.  It's just that, in this particular case, we can save ourselves work by mutating the old object, and so it's worth considering doing so...if we can do it safely/correctly.  😉

mbladel: @webbnh I very quickly wrote up a simple unit test, let me know if that's ok too - thanks again!

tests/test_upstream_issue.py
```diff
@@ -212,6 +212,40 @@ def test_github_issues_no_token(self,
         self.mock_github_repo.get_issue.assert_not_called()
         self.mock_github_issue.get_comments.assert_not_called()
 
+    @mock.patch('sync2jira.intermediary.Issue.from_github')
+    @mock.patch(PATH + 'Github')
+    @mock.patch(PATH + 'get_all_github_data')
+    def test_github_issues_no_token(self,
```
webbnh, line 218: I think this function name duplicates (and replaces??) the function defined at line 162, so it should be renamed here.

tests/test_upstream_issue.py
```diff
@@ -212,6 +212,40 @@ def test_github_issues_no_token(self,
...
+        try:
+            mock_get_all_github_data.assert_called_with(
+                'https://api.github.com/repos/org/repo/issues?labels=custom_tag%2Canother_tag%2Cand_another&filter1=filter1',
+                {'Authorization': 'token mock_token'}
+            )
+        except AssertionError:
+            mock_get_all_github_data.assert_called_with(
+                'https://api.github.com/repos/org/repo/issues?filter1=filter1&labels=custom_tag%2Canother_tag%2Cand_another',
+                {'Authorization': 'token mock_token'}
+            )
```
webbnh, line 247: If you run this test on the `main` branch code (without your changes), does it fail?  (I bet it does not....)  In that case, it's not really testing enough.

I think the test needs to call `u.github_issues()` a second time in order to generate the failure that this PR is supposed to address.

webbnh, line 247: I realize that you copied this construct from somewhere else, but I think that a different approach would be better here.  We don't actually care what the whole argument looks like, we care only that the argument contains the correct `labels` query parameter value.  So, the assertion should check just for that, in which case we don't need the `try` block or the second assertion.

So, something like this:
```suggestion
        # Assert that calls were made correctly
        expected = 'labels=custom_tag%2Canother_tag%2Cand_another'
        actual = mock_get_all_github_data.args[0]
        self.assertIn(expected, actual)
```

webbnh (changes requested):
There are a few problems with your unit test:

- [ ] It needs its own name
- [ ] It won't detect the problem that this PR is supposed to fix
- [ ] It covers only one of the two code changes you made.

tests/test_upstream_issue.py
webbnh: Also, since we are now intentionally mutating the configuration, consider adding an assertion which confirms that after the call(s):
```python
        actual = self.mock_config['sync2jira']['filters']['github']['org/repo']['labels']
        expected = 'custom_tag,another_tag,and_another'
        self.assertEqual(actual, expected)
```

tests/test_upstream_issue.py
mbladel: Oops, forgot to change it!

tests/test_upstream_issue.py
mbladel: Right, had some trouble with that but managed to make it work.

mbladel: Yep, had no idea how to go about this so just copied what was already there - so thank you very much for your suggestions, applied both.

webbnh (approved):
Looks good.  Are you planning on providing a similar change to `test_upstream_pr.py`?

mbladel:
@webbnh thanks for the review! The test for Para would basically be identical, so I don't think that's necessary.

webbnh:
> The test for Para would basically be identical, so I don't think that's necessary.

Yes it would be basically identical, but what it would be testing would be completely different!  🤣 

Yes, we don't like duplicated code, but that applies more to the product than to the tests.  Having duplicated code there is evil, too, but having duplication is better than suffering omission.

If we were being really proper, we would refactor both the issue code and the PR code and move the duplication to a common utility routine, and then we would need only a single instance of the test, because there would be only a single instance of the code.  (And, in fact, [I have a branch full of such refactorings](https://github.com/release-engineering/Sync2Jira/pull/207), but we haven't summoned the collective will to get it merged.)

webbnh (changes requested):
My suggestion to keep the mutation of the in-memory config object is probably not a good idea without additional changes.  😢 

We have [code elsewhere](https://github.com/release-engineering/Sync2Jira/blob/11123d31241dba0a393664368d19ec76664b36d1/sync2jira/upstream_issue.py#L88) (which I think you are familiar with 😉) which expects the labels to be a `list` and not a comma-separated string!  (And, Python doesn't issue an error if the argument to `isdisjoint()` is a string, because `str`'s are iterable, and it turns the string into a `set` of its individual characters, and then returns `True` if all of the labels are multi-character strings.)

So, it would probably be best to maintain the list as a `list` and only flatten it locally.  Sorry for the misdirection!  (Clearly, there is an "opportunity" here to improve our testing!)

mbladel:
Right @webbnh, so I reverted to my previous solution - hopefully that finally does it!

webbnh (approved):
It looks good.  However, I've offered a couple of coding suggestions which I think are more expressive, in case you're interested.

And, I'll put in one final plug for some testing for `sync2jira/upstream_pr.py`.  🙂

mbladel:
@webbnh feel free to push the best version of the code to my PR or even close it and open one yourself - I'm far from a Python expert but the dictionary comprehension looked more concise and worked, I'm sorry I don't really have the time to dedicate to perfecting this one little change our team needed (and I though I could contribute quickly). I appreciate the tips, though timezone difference doesn't make this easy.

Did one additional push, not sure about the test for PRs (it's exactly the same copy-pasted code), but here you go :)

webbnh (approved):
LGTM

@mbladel, thanks for your patience and for making the effort!  None of us are Python experts (and, I'm not convinced that Python experts are even useful, because "expert Python" tends to be hard to read and even harder to understand).

Everything that we do is a series of trade-offs -- the choices of code idiom, the amount of testing, how to spend our engineering time -- and ultimately, each of us has to make his/her own choice as to how those trades should go.  In terms of my code review, I'm only trying to make you aware of what those choices are (and, admittedly, to influence you in some cases 😉), but ultimately you need to make the choice yourself.  (And, if I'm unhappy with your choice, as you say, I can choose to post my own PR!  😁)

And, yeah, the timezone difference poses problems in various ways.

> the dictionary comprehension looked more concise

The problem is that conciseness isn't the only (or best) criterion.  We want the code to be easily understood by a reader while still being reasonably efficient.  To my eye, multi-line comprehensions push the boundary, and the embedded ternary compounds the problem.  (Don't get me wrong, I _love_ ternaries, but I kind of hate Python's syntax for them.)  So, for that _particular_ case, I advocate for avoiding it (particularly, since you're basically copying the the `dict` and changing one key's value, it seems like using `deepcopy()` better expresses your intention, and it _might_ be more efficient).

> not sure about the test for PRs

The code seems reasonable to me, and, more importantly, to Tox.  😀  Thanks!

mbladel:
@ralphbean ok to merge this? Thanks!

mbladel:
@ralphbean not sure why we're not merging this, is there a blocker?

PR #234, opened 2024-11-12
Remove replicated key from mock issues in tests

webbnh:
In a couple of tests, there is a key in a dictionary literal which is repeated, so the first value never gets used.  This PR removes the useless first value.


baijum (approved):
LGTM

ralphbean (approved):

Issue #171, opened 2024-07-01
Apply a specific label in non-critical error cases for team triage


mattreid:
Came up during discussion of https://github.com/release-engineering/Sync2Jira/issues/147

It could be valuable in certain situations for sync2jira to help users identify issues in bulk that may need more attention. Adding a comment on an issue helps in the moment, but doesn't provide a means to find all such issues from a JQL query/dashboard. One way this could be accomplished is with a specific label. 

As an example, I could see this helping a team rely on that label to identify a triage list of all issues that may need more attention (labels to identify the issue type were ambiguous and the issue was synced using the first value, etc). Ideally, sync2jira should be pretty hands off, so teams aren't splitting their attention between two systems frequently, but that makes it even more valuable to quickly and easily know where attention may be needed.

PR #237, opened 2024-11-16
Storypoint touchup

webbnh:
Matt and I were playing with a configuration for sync'ing priorities and story points, and it didn't seem to be working, so I dove into the code and smoothed a few rough edges.

The first commit in this branch is some small edits to the downstream-issue code to tweak the logging.  I've removed two log messages which were just repeatedly logging the configuration.  I don't think they were doing anything other than adding to the noise.  I also moved the log message which reports the creation of the downstream issue to a point where it can actually report what it is creating.  Previously, it, too, was dumping a bunch of unilluminating configuration information, and now it reports the contents of the creation request.  @ralphbean, let me know if you want any of this stuff restored.

The other commit contains a bunch of stuff:
- I added a test to filter out PR "issues" before we try to pull project information from them, because the GitHub query fails, because they are not specifically _issues_.  Please let me know if this should actually be triggering a `continue` statement to skip the whole issue, rather than skipping just the project fields.
- I tweaked a bunch of the log messages.  In some cases, I added context.  In others, I just made them more uniform.
- I did some rework of the code which locates the project node for an issue.
  - Previously, if we were unable to find a suitable project, we skipped syncing the entire issue.  Matt and I concluded that this is undesirable -- the proposed code now sync's the issue regardless of whether there is an associated project.  If there is no associated project, it doesn't try to sync the priority or storypoint fields. 
  - I reworked the handling when there is no configured project:  it should work as it did before, but the no-configured-project now has a single, combined flow.
  - These changes represent a change in policy/behavior and so they broke the unit tests.  So, I replaced the five previous unit tests with a single, matrix-based test which tests nine scenarios and matches the new behavior.
- And, I did a little code-smithing in a couple of places.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -578,12 +578,10 @@ def _get_preferred_issue_types(config, issue):
     #     'enhancement': 'Story'
     #   }
     type_list = []
-    log.debug(config)
 
     map = config['sync2jira'].get('map', {})
     conf = map.get('github', {}).get(issue.upstream, {})
 
-    log.debug(conf)
     # we consider the issue_types mapping if it exists. If it does, exclude all other logic.
     if 'issue_types' in conf:
         for tag, issue_type in conf['issue_types'].items():
@@ -616,11 +614,6 @@ def _create_jira_issue(client, issue, config):
     :returns: Returns JIRA issue that was created
     :rtype: jira.resources.Issue
     """
-    log.info("Creating %r issue for %r", issue.downstream, issue)
-    if config['sync2jira']['testing']:
-        log.info("Testing flag is true.  Skipping actual creation.")
-        return
-
     custom_fields = issue.downstream.get('custom_fields', {})
     preferred_types = _get_preferred_issue_types(config, issue)
     description = _build_description(issue)
@@ -646,7 +639,11 @@ def _create_jira_issue(client, issue, config):
     if 'labels' in issue.downstream.keys():
         kwargs['labels'] = issue.downstream['labels']
 
-    log.info("Creating issue.")
+    log.info("Creating issue for %r:  %r", issue, kwargs)
+    if config['sync2jira']['testing']:
+        log.info("Testing flag is true.  Skipping actual creation.")
+        return
+
     downstream = client.create_issue(**kwargs)
 
     # Add Epic link, QA, EXD-Service field if present
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -325,7 +325,11 @@ def github_issues(upstream, config):
 
         issue_updates = config['sync2jira']['map']['github'][upstream].get('issue_updates', {})
 
-        if 'github_project_fields' in issue_updates:
+        if '/pull/' in issue.get('html_url', ''):
+            log.debug(
+                "Issue %s/%s#%s is a pull request; skipping project fields",
+                orgname, reponame, issue['number'])
+        elif 'github_project_fields' in issue_updates:
             issue['storypoints'] = None
             issue['priority'] = ''
             issuenumber = issue['number']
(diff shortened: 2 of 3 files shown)


sync2jira/upstream_issue.py
```diff
@@ -368,25 +378,38 @@ def _get_current_project_node(upstream, project_number, issue_number, gh_issue):
...
     # If there are no project items, there is nothing to return.
     if len(project_items) == 0:
-        log.debug("Issue %s/%s is not associated with any project", upstream, issue_number)
+        log.debug("Issue %s#%s is not associated with any project",
+                  upstream, issue_number)
         return None
-    # If there is exactly one project item, return it if we don't have a configured project.
-    if not project_number and len(project_items) == 1:
+    # If there is exactly one project item, return it.  Even if it doesn't
+    # match the configured project, having one is better than having none.
```
baijum, line 385: If the user has explicitly specified a project, it is better to stick to that project for syncing. And if an issue is linked to some other project, that project may not be having the required fields, and the expected format.

sync2jira/downstream_issue.py
```diff
@@ -646,7 +639,11 @@ def _create_jira_issue(client, issue, config):
     if 'labels' in issue.downstream.keys():
         kwargs['labels'] = issue.downstream['labels']
 
-    log.info("Creating issue.")
+    log.info("Creating issue for %r:  %r", issue, kwargs)
```
baijum, line 642: This is not really a storypoints PR related change. Since it's a very small change to improve logging, I am okay with this change.

sync2jira/upstream_issue.py
```diff
@@ -320,37 +320,47 @@ def github_issues(upstream, config):
 
         issue_updates = config['sync2jira']['map']['github'][upstream].get('issue_updates', {})
 
-        if 'github_project_fields' in issue_updates:
+        if '/pull/' in issue.get('html_url', ''):
```
baijum, line 323: Why do we need to exclude PRs? GitHub allows a PR to be added to a project.

sync2jira/upstream_issue.py
webbnh: If `sync2jira` has been requested to sync a given issue (either via a fedmsg notification or because we're initializing a repo) and the issue is not associated with the project specified in the configuration, we have three options:

1. ignore the issue altogether
2. sync only the non-project fields of the issue
3. try to sync all the fields and ignore only the individual fields which don't "work"

The existing code uses the first option, and that resulted in a "user surprise" for me and Matt.  It seems like the tool should follow one of the latter two options.

I realize that sync'ing priorities might be problematic, because the upstream values are project-specific keywords and the downstream values are project-schema-dependent, but I believe the code is already equipped to handle the case where the configuration doesn't match the source or destination, so I figured it wasn't really a problem to try to sync the priorities -- at worst the downstream simply wouldn't update that field.

As for story points, they are just numbers, so, as long as we can source the value, we can update it.  However, the fact that it comes from the "wrong" project is arguably a good reason against trying to sync it, but I concluded that bad data was better than no data.  (There is a problem with visibility:  how does a developer lookin in Jira know whether it's in sync with the upstream?...I was hoping that bad data would be more obvious and motivating than no data...but, perhaps that's wrong.)

In any case, I think we should move away from option 1, but I'm happy to go to option 2 instead of option 3.  Thoughts?

sync2jira/upstream_issue.py
webbnh: Without this exclusion, the error reported at (original) line 336 gets triggered, because the PR is not an issue as far as the `ghquery` is concerned.

Also, there's an entirely separate code flow in [`sync2jira/upstream_pr.py`](https://github.com/release-engineering/Sync2Jira/blob/main/sync2jira/upstream_pr.py) which is ostensibly for handling PRs -- the code here is just for proper Issues.

baijum: Ok!

sync2jira/upstream_issue.py
mattreid: I think lack of data is more noticeable than wrong data, especially for something that we're hoping is fairly hands off for teams once they get things set up, but I'm happy with anything that isn't option 1 :) It would be easier to see at the aggregate level this data will be used for that there is data missing from synced issues, than to recognize the nuance of incorrectly synced data that pulled from an unexpected source. As you say though, there's no guarantee that the values would match between projects, so if we handle that gracefully, there may not be much difference between the two approaches after the issue is synced using any data that maps correctly.

I've seen data from our teams that shows not all issues in a repo are always included in the related project, so ignoring those issues altogether was a problem. Since sync2jira is mapped at the repository level and not a project, it makes sense to me that we'd sync using the base data we have available. I haven't seen off-hand that an issue is part of more than 1 project, but it could be possible. It could also be possible to have 1 project association for an issue, but have it not match the configured project number, although I don't remember seeing an example of this yet.

https://github.com/release-engineering/Sync2Jira/issues/171 could potentially help us identify ambiguous issues in the future, if someone decides to pick it up, by adding some mechanism that indicates some further triage may be needed.

webbnh: > there may not be much difference between the two approaches after the issue is synced using any data that maps correctly.

Except for the fact that it (i.e., storypoints) would be coming from the wrong source and therefore would quite possibly be wrong....

> I haven't seen off-hand that an issue is part of more than 1 project, but it could be possible. It could also be possible to have 1 project association for an issue, but have it not match the configured project number, although I don't remember seeing an example of this yet.

If you're really interested, I should be able to dig up examples of both.  There is at least one instance of the second in the `intellij-common` repo configuration that you and I were testing with.  And, I've hit instances of the first in the `eclipse-che` organization.

> https://github.com/release-engineering/Sync2Jira/issues/171 could potentially help us

Yes:  it would be great if there were some way to make problems visible (e.g., via Jira) to the end-users, who don't have ready access to the `sync2jira` logs.

I'll modify the code to use Option 2.

webbnh:
This PR is now ready for review.

baijum (approved):
LGTM

ralphbean (approved):

PR #239, opened 2024-11-26
Workaround for FedMsg storypoint (str) values

webbnh:
Attempting to actually use the storypoint sync'ing [fails](https://issues.redhat.com/browse/IDEEXT-366?focusedId=26153477&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26153477) because the storypoint value must be a number, and we're (apparently) presenting it as a string.

The problem arises from the fact that, until we merge #207, there is substantial duplication in the source, and the change in #216 to introduce support for storypoint sync'ing covered only the repo-initialization path and not the FedMsg update path.

This PR introduces a stop-gap into the downstream issue code which attempts to force the story point value to be an `int`, and, if it fails, simply skips the update.  Once the upstream/intermediate code is corrected to supply a numeric value (or no value), this conversion can be removed.

This PR also tweaks the downstream code for setting Priority values and improves the logging of errors, and it adds some unit tests to detect these problems.

CC: @baijum

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -979,6 +979,14 @@ def _update_github_project_fields(client, existing, issue,
     for name, values in github_project_fields.items():
         fieldvalue = getattr(issue, name)
         if name == 'storypoints':
+            if not isinstance(fieldvalue, int):
+                # FIXME:  The intermediate issue _should_ have either a number
+                #  or None for this value, but the code isn't complete yet; so,
+                #  until that is addressed, cover for it here.
+                try:
+                    fieldvalue = int(fieldvalue)
+                except (TypeError, ValueError):
+                    continue
             try:
                 jirafieldname = default_jira_fields['storypoints']
             except KeyError:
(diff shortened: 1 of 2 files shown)


mattreid:
As part of 

> This PR introduces a stop-gap into the downstream issue code which attempts to force the story point value to be an int, and, if it fails, simply skips the update.

Does that mean it would still log any error that it didn't match/wasn't able to convert (by commenting on the downstream issue), or it truly skips and we don't get anything indicating that something may have been missed?

tests/test_downstream_issue.py
```diff
@@ -1709,3 +1722,27 @@ def test_update_github_project_fields_priority(self, mock_client):
...
+                "gh_field": "Priority",
+                "options": {
+                    "P0": "Blocker",
+                    "P1": "Critical",
+                    "P2": "Major",
+                    "P3": "Minor",
+                    "P4": "Optional",
+                    "P5": "Trivial"
+                }}}
+        for bad_pv in [None, '', 'bad_value']:
```
mattreid, line 1742: Bad value in this case means an unexpected mapping (either GH Project value or Jira value that don't exist in the field list)? If it doesn't have a value at all (does that show up as Null? Empty?), we skip that field by default?

webbnh: I believe that `'bad_value'` should be slotted in where we would expect to find, e.g., `'P5'`, and it would result in a failure-to-map result, which I think is analogous to `None` (but it could be `''`) -- that is, all three values should have the same result, but they represent different initial conditions -- and they each should result in skipping the attempt to update the downstream issue.

webbnh:
> As part of
> 
> > This PR introduces a stop-gap into the downstream issue code which attempts to force the story point value to be an int, and, if it fails, simply skips the update.
> 
> Does that mean it would still log any error that it didn't match/wasn't able to convert (by commenting on the downstream issue), or it truly skips and we don't get anything indicating that something may have been missed?

For the case of story points, it will silently skip the update, and no one will be the wiser..._if_ the input value cannot be coerced to an integer.  I think that this is OK, since story points should be numeric, and the most likely reason why it cannot be converted is that it is missing (`null`, `None`, or an empty string), in which case we _want_ to ignore it.

For the case of priorities, we were already silently skipping the update for unmappable values, so this doesn't make anything worse.  Nevertheless, I concur that the lack of feedback to the maintainers is disappointing.  But, balanced against that is the fact that sending errors to the log is not a reliable way to get them noticed, and adding them as comments becomes noisy if they get re-added every time the issue is touched.  So, I'm not sure what the best solution here is.

But, this PR is not the "final fix".  We can and should implement that after we've DRYed out the code...not before.

sync2jira/downstream_issue.py
```diff
@@ -979,6 +979,14 @@ def _update_github_project_fields(client, existing, issue,
     for name, values in github_project_fields.items():
         fieldvalue = getattr(issue, name)
         if name == 'storypoints':
+            if not isinstance(fieldvalue, int):
```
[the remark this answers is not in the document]
webbnh: You are technically correct -- if `fieldvalue` is an `int`, then the "conversion" will proceed without error, and we'll end up in the same place as we do if we skip the `try` block -- however, my hope is that, once the upstream/intermediate-issue code is corrected, we will remove the `try` block and simply `continue` if `fieldvalue` is not an `int` (because it will be `None` in that case or otherwise _supposed_ to be excluded), so we'll need this check back at that point, so we might as well include it now.

PR #240, opened 2024-12-02
Add diagnostics for GitHub Project fields

webbnh:
Trying to diagnose why story points don't show up in Jira (and don't produce any errors, either).


mattreid (approved):
Curious to see what this surfaces. +1

PR #254, opened 2024-12-05
Replace string interpolation with log arguments and f-strings

webbnh:
This PR is the next in the series of refactorings originally set out in https://github.com/release-engineering/Sync2Jira/pull/207 which endeavors to replace duplicated code in Sync2Jira with common subroutines; this is the follow-on to https://github.com/release-engineering/Sync2Jira/pull/253.

These changes center on removing the uses of Python2-style string interpolation, replacing them with more appropriate operations.

For inserting values into log messages, the logger will do the interpolation automatically if the values are provided as arguments to the call.  This approach has the added advantage that the interpolation is deferred until the message is actually logged, such that, if the intended class of logging is disabled (e.g., calling `log.debug()` when the log level is set to "INFO") the interpolation will be skipped, since the message won't be logged, which is a performance savings.  So, this PR removes uses of the interpolation operator (`%`) from logging calls instead passing the to-be-interpolated values as arguments to the logger call.

For inserting values into other strings, Python3 offers "f-strings" which are generally more expressive and easier to read than the Python2 format arguments used with the interpolation operator.  So, this PR makes that substitution where it improves the readability of the code.

For converting values to strings without further augmentation, this PR replaces the interpolation with a call to `str()`.

Added code to trim whitespace from the issue owner, iff the owner is not `None`, blank, or in some other way absent or unpleasant.  (A similar change for the admin username is coming in a subsequent commit.)

And, removed a couple of unneeded list instantiations.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -109,8 +109,8 @@ def get_jira_client(issue, config):
 
     if not isinstance(issue, Issue) and not isinstance(issue, PR):
         log.error("passed in issue is not an Issue instance")
-        log.error("It is a %s" % type(issue).__name__)
-        raise TypeError("Got %s, expected Issue" % type(issue).__name__)
+        log.error("It is a %s", type(issue).__name__)
+        raise TypeError(f"Got {type(issue).__name__}, expected Issue")
 
     # Use the Jira instance set in the issue config. If none then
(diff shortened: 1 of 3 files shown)


ralphbean (approved):

ralphbean:
Makes total sense. :) Thanks for cranking on this @webbnh.

PR #255, opened 2024-12-06
Use iterators instead of instantiating lists/tuples

webbnh:
This PR is the next in the series of refactorings originally set out in https://github.com/release-engineering/Sync2Jira/pull/207 which endeavors to improve code quality in Sync2Jira; this is the follow-on to https://github.com/release-engineering/Sync2Jira/pull/254.

This change replaces instances in the code where we instantiate a list from a generator only to access the first element.  Instead, we use the `next()` function to produce only the first element and avoid the list instantiation and the overhead of repeatedly invoking the generator.


ralphbean (approved):
Nice and clean!

PR #256, opened 2024-12-06
Simplify PR #222

webbnh:
This PR is the next in the series of refactorings originally set out in https://github.com/release-engineering/Sync2Jira/pull/207 which endeavors to improve code quality in Sync2Jira; this is the follow-on to https://github.com/release-engineering/Sync2Jira/pull/255.

This change is a refinement of #222 which utilizes the simpler timezone support included in the `datetime` package rather than adding the `ZoneInfo` package, and it simplifies the instantiations of timestamps in the unit tests.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -17,18 +17,17 @@
 #
 # Authors:  Ralph Bean <rbean@redhat.com>
 
-# Build-In Modules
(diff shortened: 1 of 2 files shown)


ralphbean:
This is the first scary one. :)

ralphbean (approved):

PR #259, opened 2024-12-07
Pick lint

webbnh:
This PR is the next in the series of refactorings originally set out in https://github.com/release-engineering/Sync2Jira/pull/207 which endeavors to improve code quality in Sync2Jira; this is the follow-on to https://github.com/release-engineering/Sync2Jira/pull/258.

This change picks a bunch of lint from a few of the modules to make the code "more Pythonic":
- Change `camelCase` names to `snake_case`.
- The first parameter to a class method is conventionally named `cls` and should be used in preference to referencing the class by name.
- Correct the reference to `fedmsg.config.load_config`.
- Replace cryptic/abbreviated variable names when there is no reason not to spell them out.
- Correct misspelled words in comments and test names; wordsmith a couple of comments.

### DIFF
--- a/sync2jira/intermediary.py
+++ b/sync2jira/intermediary.py
@@ -36,7 +36,7 @@ class Issue(object):
         self.storypoints = storypoints
 
         # First trim the size of the content
-        self.content = trimString(content)
+        self.content = trim_string(content)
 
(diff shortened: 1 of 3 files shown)


ralphbean (approved):
Straightforward! Thank you.

PR #267, opened 2024-12-11
Promote `add_project_values()` error logging from DEBUG to INFO

webbnh:
This PR follows #263 and #247:  it's not evident, yet, why the story point values aren't being updated.  There is some evidence that the attempt to fetch them from the GitHub Project is failing, but these failures aren't included in the log; so this PR promotes the logging in `add_project_values()` from `DEBUG` to `INFO` so that the log will show when there are errors accessing the GitHub Project.

This PR also slightly refactors `add_project_values()`, removing the nested levels of indentation by simply `return`'ing at the points where we determine that we cannot or will not update the GH project values.  (I suggest suppressing whitespace differences when reviewing the changes.)

### DIFF
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -209,52 +209,56 @@ def add_project_values(issue, upstream, headers, config):
     :param dict headers: HTTP Request headers, including access token, if any
     :param dict config: Config
     """
-    orgname, reponame = upstream.rsplit('/', 1)
     upstream_config = config['sync2jira']['map']['github'][upstream]
     project_number = upstream_config.get('github_project_number')
     issue_updates = upstream_config.get('issue_updates', {})
-    if 'github_project_fields' in issue_updates:
-        issue['storypoints'] = None
-        issue['priority'] = None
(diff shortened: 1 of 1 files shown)


ralphbean (approved):
Makes sense.

(nit: the diff is 99% about the refactor while the PR and commit suggests it is only about a logging change.)

webbnh:
> the diff is 99% about the refactor while the PR

If you suppress the whitespace changes, then it's more like 62.5%, but I take your point.  🙂 

> commit suggests it is only about a logging change.

If you're looking [here on the website](https://github.com/release-engineering/Sync2Jira/pull/267/commits), you have to click on the ellipsis to see the whole log message (it all shows [here](https://github.com/release-engineering/Sync2Jira/commit/b9069c61e771cbb90bebd31953fd9def7a689d4d), though).

Still, the commit was motivated by the logging change which _should_ be the only visible/functional change in the behavior.  So, I put the important stuff in the first 72 characters, and left the rest to the next line of the commit log message (and, yeah, Git and related tools treat the first line as a summary and tend to elide the rest of the log).

PR #277, opened 2025-01-24
Handle other PR events

ralphbean:
PRs can be labelled and "milestone'd" as well as just commented on, which get exposed issue events. Here, broaden the terms of this condition to cover all of those events.

### DIFF
--- a/sync2jira/main.py
+++ b/sync2jira/main.py
@@ -289,9 +289,9 @@ def handle_msg(msg, suffix, config):
     """
     issue = None
     pr = None
-    # GitHub '.issue.' is used for both PR and Issue
+    # GitHub '.issue*' is used for both PR and Issue
     # Check for that edge case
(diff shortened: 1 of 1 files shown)


webbnh (approved):
This change looks OK to me, although Black will not be happy with it, if #280 gets merged first.

However, in order for this to work properly, I believe that this branch/PR should include [commit `efb0a01`](https://github.com/release-engineering/Sync2Jira/commit/efb0a0198f8cda1f11d52bc6ea2b7070d66761ed), which is currently in #280 but which should really be in this PR **instead**.

Also, we need to consider whether this change as presented is sufficient, or whether we need to tweak [the condition](https://github.com/release-engineering/Sync2Jira/blob/d841d31d4dc86e8c111f1464f2802e8e69d52d11/sync2jira/upstream_issue.py#L135) in `handle_github_message()`.

PR #278, opened 2025-01-24
Update to python 3.13

ralphbean:


webbnh (approved):
LGTM, although I have a couple of questions.

Do you really want to _replace_ 3.12 with 3.13, or should we be _adding_ 3.13 and keeping 3.12?  On a related note, why drop 3.12 and not _also_ drop 3.9?  (I.e., what is the requirement that we're meeting, here?)

ralphbean:
Yeah, I had a similar thought (to just drop 3.9). Our only real use is in our deployment and I think that's on 3.13.

PR #279, opened 2025-01-24
lint: apply isort to standardize imports

ralphbean:


webbnh (dismissed):
LGTM, but I have two questions/thoughts.

First, are you sure that you want to run `isort` as a separate Tox "environment", rather than running it as an additional command in the `lint` environment?

Second, I'm a fan of `isort`, and I recommend a few more options:
```
known_first_party = ["sync2jira"]  # use separate section for project sources
force_sort_within_sections = true  # don't separate import vs from
order_by_type = false              # sort alphabetic regardless of case
```
(Notes:  Don't include the `#` comments -- `isort` cannot read the configuration successfully with them present in the `tox.ini` file (although, they are OK in a `pyproject.toml` file -- different rules, I guess).  Also, these options will change the sort orders, so you'll need to re-generate the changes.)

sync2jira/downstream_issue.py
```diff
@@ -17,8 +17,9 @@
 #
 # Authors:  Ralph Bean <rbean@redhat.com>
 
-# Python Standard Library Modules
 from datetime import datetime, timezone
+
+# Python Standard Library Modules
 import difflib
 import logging
 import operator
```
webbnh, line 25: These imports should all be in one block (i.e., `datetime` _is_ a Standard Library module), so we should probably just remove the comment at line 22 (and line 29 and line 36) -- now that we have `isort`, these blocks will be maintained automatically, no there is no need to label them.

sync2jira/main.py
```diff
@@ -21,28 +21,29 @@
...
 import logging
-import warnings
-import traceback
-from time import sleep
-import requests
-from copy import deepcopy
 import os
+from time import sleep
+import traceback
+import warnings
```
webbnh, line 31: Ditto previous comment.

webbnh (commented):
Looks like we collided.  Anyway, here are some late-breaking comments, if you're interested.

PR #280, opened 2025-01-24
Format python with black

ralphbean:
Just for consistent style.


ralphbean:
This includes changes from https://github.com/release-engineering/Sync2Jira/pull/279.

tests/integration_tests/integration_test.py
```diff
@@ -129,15 +148,17 @@ def get_jira_client():
...
-if __name__ == '__main__':
+    return jira.client.JIRA(
+        **{
+            "options": {
+                "server": URL,
+                "verify": False,
+            },
+            "token_auth": TOKEN,
+        }
+    )
```
webbnh, line 159: Using the `**` operator on a literal is a little heavy-handed, we should do something like this, instead:
```suggestion
    return jira.client.JIRA(
        options={
            "server": URL,
            "verify": False,
        },
        token_auth=TOKEN,
    )
```
_[Again, feel free to decline if you aren't interested in this kind of change in this PR.]_

tests/test_intermediary.py
```diff
@@ -3,199 +3,204 @@
...
-            'sync2jira': {
-                'map': {
-                    'github': {
-                        'github': {'mock_downstream': 'mock_key'}
-                    }
-                }
+            "sync2jira": {
+                "map": {"github": {"github": {"mock_downstream": "mock_key"}}}
             }
         }
```
webbnh, line 19: Here's a place where I would have used Black's "magic comma" to preserve the old formatting:
```suggestion
        self.mock_config = {
            'sync2jira': {
                'map': {
                    'github': {
                        'github': {'mock_downstream': 'mock_key'}
                    },
                },
            },
        }
```

webbnh (approved):
I'm a fan of using Black, even if I don't necessarily like everything that it does.

I've called out a few things below:
- Foremost, it looks like you included a commit on this branch which belongs in a separate PR.
- Beyond that, there are a couple of changes that Black made that I think you might want to resist in various ways.
- And, finally, the changes highlighted a couple of spots in the code where we might want to make some tweaks beyond what Black proposed...which you should feel free to decline.

sync2jira/main.py
```diff
@@ -61,6 +61,10 @@
 # Issue related handlers
 issue_handlers = {
     # GitHub
+    # New webhook-2fm topics
+    'github.issues': u_issue.handle_github_message,
+    'github.issue_comment': u_issue.handle_github_message,
+    # Old github2fedmsg topics
```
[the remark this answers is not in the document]
ralphbean: oop! you're right!

docs/source/conf.py
```diff
@@ -119,15 +120,12 @@
...
     #
     # 'pointsize': '10pt',
-
     # Additional stuff for the LaTeX preamble.
     #
     # 'preamble': '',
-
     # Latex figure (float) alignment
     #
     # 'figure_align': 'htbp',
```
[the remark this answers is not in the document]
ralphbean: Hm, fair enough. I added a `#` to keep it consistent.

tests/test_intermediary.py
ralphbean: I'll let this one go. The reason I love black is that I get to stop thinking about things like this and focus on other things :)

PR #281, opened 2025-01-24
Support webhook-2fm style messages

ralphbean:
https://lists.fedoraproject.org/archives/list/infrastructure@lists.fedoraproject.org/message/UGMBWWAXIM6BHK3RIBWQWCNIUYZ6U5XH/

I found some example messages here (notice the different topic):

* https://apps.fedoraproject.org/datagrepper/v2/search?rows_per_page=20&delta=127800&topic=org.fedoraproject.prod.github.issues
* https://apps.fedoraproject.org/datagrepper/v2/search?rows_per_page=20&delta=127800&topic=org.fedoraproject.prod.github.issue_comment

### DIFF
--- a/sync2jira/compat.py
+++ b/sync2jira/compat.py
@@ -4,11 +4,9 @@ def extract_message_body(msg):
     Handle both fedmsg and fedora-messaging style message bodies.
     """
 
-    if "body" in msg:
-        return msg["body"]
-    elif "msg" in msg:
-        return msg["msg"]
-    else:
-        raise KeyError(
-            f"Unrecognized message format with keys {msg.keys()}. Expected either 'msg' or 'body'"
-        )
+    body = msg.get("body", msg.get("msg"))
+    if body:
+        return body
+    raise KeyError(
+        f"Unrecognized message format with keys {msg.keys()}. Expected either 'msg' or 'body'"
+    )
--- a/sync2jira/compat.py
+++ b/sync2jira/compat.py
@@ -1,0 +1,14 @@
+def extract_message_body(msg):
+    """Strip off the message envelope.
+
+    Handle both fedmsg and fedora-messaging style message bodies.
+    """
+
+    if "body" in msg:
+        return msg["body"]
+    elif "msg" in msg:
+        return msg["msg"]
+    else:
+        raise KeyError(
+            f"Unrecognized message format with keys {msg.keys()}. Expected either 'msg' or 'body'"
+        )
--- a/sync2jira/main.py
+++ b/sync2jira/main.py
@@ -38,6 +38,7 @@ import requests
 from requests_kerberos import HTTPKerberosAuth, OPTIONAL
 
 # Local Modules
+import sync2jira.compat as c
 import sync2jira.downstream_issue as d_issue
 import sync2jira.downstream_pr as d_pr
 from sync2jira.intermediary import matcher
@@ -62,6 +63,10 @@ failure_email_subject = "Sync2Jira Has Failed!"
 # Issue related handlers
 issue_handlers = {
     # GitHub
+    # New webhook-2fm topics
+    "github.issues": u_issue.handle_github_message,
+    "github.issue_comment": u_issue.handle_github_message,
+    # Old github2fedmsg topics
     "github.issue.opened": u_issue.handle_github_message,
     "github.issue.reopened": u_issue.handle_github_message,
     "github.issue.labeled": u_issue.handle_github_message,
@@ -78,6 +83,10 @@ issue_handlers = {
 # PR related handlers
 pr_handlers = {
     # GitHub
+    # New webhook-2fm topics
+    "github.pull_request": u_pr.handle_github_message,
+    "github.issue_comment": u_pr.handle_github_message,
+    # Old github2fedmsg topics
     "github.pull_request.opened": u_pr.handle_github_message,
     "github.pull_request.edited": u_pr.handle_github_message,
     "github.issue.comment": u_pr.handle_github_message,
@@ -282,8 +291,8 @@ def initialize_recent(config):
 
         # Deal with the message
         log.debug("Handling %r %r", suffix, entry["topic"])
-        msg = entry["msg"]
-        handle_msg({"msg": msg}, suffix, config)
+        body = c.extract_message_body(entry)
+        handle_msg({"body": body}, suffix, config)
 
 
 def handle_msg(msg, suffix, config):
@@ -298,7 +307,8 @@ def handle_msg(msg, suffix, config):
     # GitHub '.issue*' is used for both PR and Issue
     # Check for that edge case
     if suffix.startswith("github.issue"):
-        if "pull_request" in msg["msg"]["issue"] and msg["msg"]["action"] != "deleted":
+        body = c.extract_message_body(msg)
+        if "pull_request" in body["issue"] and body["action"] != "deleted":
             # pr_filter turns on/off the filtering of PRs
             pr = issue_handlers[suffix](msg, config, pr_filter=False)
             if not pr:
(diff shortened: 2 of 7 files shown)


sync2jira/compat.py
```diff
@@ -0,0 +1,14 @@
...
+
+    Handle both fedmsg and fedora-messaging style message bodies.
+    """
+
+    if 'body' in msg:
+        return msg['body']
+    elif 'msg' in msg:
+        return msg['msg']
+    else:
+        raise KeyError(f"Unrecognized message format with keys {msg.keys()}. Expected either 'msg' or 'body'")
```
webbnh, line 14: Alternatively,
```suggestion
    body = msg.get("body", msg.get("msg"))
    if body:
        return body
    raise KeyError(f"Unrecognized message format with keys {msg.keys()}. Expected either 'msg' or 'body'")
```
This avoids duplicate key look-ups, and it also raises an error if the value exists but is empty.  (If you don't like this last part, you can change the `if` test to check for `None` explicitly to allow it to return an empty body.)

sync2jira/main.py
```diff
@@ -292,7 +301,8 @@ def handle_msg(msg, suffix, config):
     # GitHub '.issue.' is used for both PR and Issue
     # Check for that edge case
     if suffix == 'github.issue.comment':
-        if 'pull_request' in msg['msg']['issue'] and msg['msg']['action'] != 'deleted':
+        body = c.extract_message_body(msg)
```
webbnh, line 304: Rather than adding the call to `extract_message_body()` here, consider changing the function signature to receive the body directly instead of the message wrapper for it.  This will save the dict instantiation at line 289 at the possible cost of having to add a call to `extract_message_body()` in `listen()` at line 172 (but, since you would be removing the call from here, that would be a net-zero cost and an overall savings, at the cost of an extra line of code change).

sync2jira/upstream_issue.py
```diff
@@ -121,8 +122,9 @@ def handle_github_message(msg, config, pr_filter=True):
     :returns: Issue object
     :rtype: sync2jira.intermediary.Issue
     """
-    owner = msg['msg']['repository']['owner']['login']
-    repo = msg['msg']['repository']['name']
+    body = c.extract_message_body(msg)
```
webbnh, line 125: Ditto the previous comment.

sync2jira/upstream_pr.py
```diff
@@ -38,8 +39,9 @@ def handle_github_message(msg, config, suffix):
     :returns: Issue object
     :rtype: sync2jira.intermediary.PR
     """
-    owner = msg['msg']['repository']['owner']['login']
-    repo = msg['msg']['repository']['name']
+    body = c.extract_message_body(msg)
```
webbnh, line 42: Ditto the previous comment.

tests/test_upstream_issue.py
```diff
@@ -63,6 +63,29 @@ def setUp(self):
                 }
             }
         }
+        self.new_style_mock_github_message = {
+            'body': {
```
webbnh, line 67: Is `new_style_mock_github_message` supposed to be identical to `old_style_mock_github_message` except for the `body` vs. `msg` key?  If so, I recommend defining the body separately and using that definition to define these two attributes (or to define one attribute using the other), to avoid repeating ourself.

tests/test_upstream_issue.py
```diff
@@ -564,7 +587,42 @@ def test_handle_github_message_no_comments(self, mock_issue_from_github, mock_gi
...
+                                                   'filter1': 'filter1', 'user':
+                                                       {'login': 'mock_login', 'fullname': 'mock_name'},
+                                                   'milestone': 'mock_milestone'}, self.mock_config)
+        mock_github.assert_called_with('mock_token', retry=5)
+        self.assertEqual('Successful Call!', response)
+        self.mock_github_client.get_repo.assert_called_with('org/repo')
+        self.mock_github_repo.get_issue.assert_called_with(number='mock_number')
+        self.mock_github_issue.get_comments.assert_any_call()
+        self.mock_github_client.get_user.assert_called_with('mock_login')
+
```
webbnh, line 622: Rather than cookie-cuttering this code, I suggest reusing the code in `test_handle_new_style_github_message_successful()` with each of the old and new style of message, to avoid the code duplication.

webbnh (approved):
Looks good, but I've proposed some improvements for your consideration.

sync2jira/compat.py
ralphbean: This is a level of optimization I almost never think about. Thank you for the suggestion!

tests/test_upstream_issue.py
ralphbean: You got it. Done in a305ae3.

sync2jira/upstream_pr.py
ralphbean: Done in a305ae3.

sync2jira/upstream_issue.py
ralphbean: Done in a305ae3.

sync2jira/main.py
ralphbean: Done in a305ae3.

tests/test_upstream_issue.py
ralphbean: After a305ae3, this can be done more simply.

ralphbean: Like in 2218859 ?

tests/test_main.py
```diff
@@ -30,7 +30,15 @@ def setUp(self):
...
-        self.mock_message = {"msg_id": "mock_id", "msg": {"issue": "mock_issue"}}
+        self.mock_message_body = {"issue": "mock_issue"}
+        self.old_style_mock_message = {
+            "msg_id": "mock_id",
+            "msg": self.mock_message_body,
+        }
+        self.new_style_mock_message = {
+            "msg_id": "mock_id",
+            "msg": self.mock_message_body,
+        }
```
webbnh, line 41: The key in the new style message (line 40) is supposed to be `body`, right?

sync2jira/main.py
```diff
@@ -282,14 +292,14 @@ def initialize_recent(config):
...
         # Deal with the message
         log.debug("Handling %r %r", suffix, entry["topic"])
-        msg = entry["msg"]
-        handle_msg({"msg": msg}, suffix, config)
+        body = c.extract_message_body(entry)
+        handle_msg({"body": body}, suffix, config)
 
 
-def handle_msg(msg, suffix, config):
+def handle_msg(body, suffix, config):
```
webbnh, line 299: `handle_msg()` now receives the unwrapped message body, right?  So line 296 should just pass `body` as the first parameter, without the wrapping, similar to line 177, right?

webbnh (changes requested):
It looks great!  But, unfortunately, I think there are a few things you need to touch up before you merge.

tests/test_main.py
ralphbean: You're right! Hasty, hasty.

sync2jira/main.py
ralphbean: You're right! Fixing.

tests/test_main.py
```diff
@@ -426,7 +463,7 @@ def test_initialize_recent(self, mock_query, mock_handle_msg):
 
         # Assert everything was called correctly
         mock_handle_msg.assert_called_with(
-            {"msg": "mock_msg"}, "github.issue.comment", self.mock_config
+            {"body": "mock_msg"}, "github.issue.comment", self.mock_config
         )
```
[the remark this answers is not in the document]
ralphbean: +1

webbnh (approved):
LGTM

PR #282, opened 2025-01-25
Tweak handling PR messages

webbnh:
This change reworks the dispatching code in `handle_msg()` removing the preamble special case to generalize the code and ensure that it doesn't miss the existence checks for the message handlers.

It moves the special case which handles certain PR messages as Issues and embeds it into the Issues case.  That is, it reduces the flow to two cases -- Issues and PRs -- checking first to see if the message suffix matches an Issue handler and, if not, then checking if it matches a PR handler.  And, it untangles the two flows so that each reaches its respective completion before returning, instead of coming back together for subsequent conditionals.

If the message suffix matches an Issue handler, _then_ we determine whether the message is actually referencing a PR; otherwise, the sequence of operations is substantially the same as the previous code, with fewer conditionals and state variables.

This change also reworks the upstream issue `handle_github_message()` function's third parameter, replacing the previous "filter request" control with a boolean declaring whether the message is known to reference a PR instead of an Issue.  This allows us to collapse into one test the two discrete configuration checks of whether Issues and PRs, respectively, are being sync'd.  And, it simplifies another check which skips all PRs which are not being closed.

Unfortunately, I also had to tweak two of the unit tests whose mocking of the `issue_handlers` dictionary didn't properly handle using the `get()` method it.

### DIFF
--- a/sync2jira/main.py
+++ b/sync2jira/main.py
@@ -303,33 +303,29 @@ def handle_msg(body, suffix, config):
     :param String suffix: Incoming suffix
     :param Dict config: Config dict
     """
-    issue = None
-    pr = None
-    # GitHub '.issue*' is used for both PR and Issue
-    # Check for that edge case
-    if suffix.startswith("github.issue"):
-        if "pull_request" in body["issue"] and body["action"] != "deleted":
-            # pr_filter turns on/off the filtering of PRs
-            pr = issue_handlers[suffix](body, config, pr_filter=False)
-            if not pr:
+    if handler := issue_handlers.get(suffix):
+        # GitHub '.issue*' is used for both PR and Issue; check if this update
+        # is actually for a PR
+        if "pull_request" in body["issue"]:
+            if body["action"] == "deleted":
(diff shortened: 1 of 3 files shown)


webbnh:
Looking at the log for other reasons, I just spotted another indication of the leak which this change is intended to fix:
```
[2025-01-30 17:48:50,948] INFO: GitHub error while fetching issue redhat-developer/lsp4ij#786: {"data":{"repository":{"issue":null}},"errors":[{"type":"NOT_FOUND","path":["repository","issue"],"locations":[{"line":6,"column":11}],"message":"Could not resolve to an Issue with the number of 786."}]}
[2025-01-30 17:48:50,950] INFO: [PR] Considering upstream https://github.com/redhat-developer/lsp4ij/pull/786, [redhat-developer/lsp4ij] feat: Inlay hints refactor
[2025-01-30 17:48:50,950] INFO: [PR] No match found for [redhat-developer/lsp4ij] feat: Inlay hints refactor
```
We're calling `add_project_values()` for something which we expect to be an Issue which is in fact a PR.

ralphbean (approved):

PR #287, opened 2025-02-03
Fix fedora-messaging schema handling

ralphbean:
There's a "msg" presented at the top layer, and then a "body" presented inside that.


webbnh (approved):
LGTM

PR #289, opened 2025-02-04
Port to fedora-messaging

ralphbean:
Fedora infrastructure is going to decommission the fedmsg-gateway endpoint we've been consuming, which requires us to migrate off that stack entirely.


ralphbean:
In order to get this working locally I had to install the `fedora-messaging` rpm and edit the `/etc/fedora-messaging/config.toml file to set the `amqp_url` to `"amqps://fedora:@rabbitmq.fedoraproject.org/%2Fpublic_pubsub"`.

tests/test_main.py
```diff
@@ -289,67 +292,57 @@ def test_initialize_github_error(
...
         # Assert everything was called correctly
         mock_handle_msg.assert_not_called()
 
     @mock.patch(PATH + "handle_msg")
     @mock.patch(PATH + "issue_handlers")
-    @mock.patch(PATH + "fedmsg")
-    def test_listen_no_issue(self, mock_fedmsg, mock_handlers_issue, mock_handle_msg):
+    @mock.patch(PATH + "load_config")
+    def test_listen_no_issue(
+        self, mock_load_config, mock_handlers_issue, mock_handle_msg
```
webbnh, line 314: `issue_handlers` is a `dict` -- we should probably be using [`@mock.patch.dict()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict) to patch it.  (See lines [401-403](https://github.com/release-engineering/Sync2Jira/blob/fb8ca0eb9447fedc7f0fa90976fe4177fab591da/tests/test_main.py#L401-L403) and [421-423](https://github.com/release-engineering/Sync2Jira/blob/fb8ca0eb9447fedc7f0fa90976fe4177fab591da/tests/test_main.py#L421-L423).)

tests/test_main.py
```diff
@@ -289,67 +292,57 @@ def test_initialize_github_error(
...
         # Set up return values
         mock_handlers_issue["github.issue.comment"].return_value = None
-        mock_fedmsg.tail_messages.return_value = [
-            ("dummy", "dummy", "d.d.d.github.issue.drop", self.old_style_mock_message)
-        ]
+        mock_load_config.return_value = self.mock_config
 
         # Call the function
-        m.listen(self.mock_config)
+        self.old_style_mock_message.topic = "d.d.d.github.issue.drop"
```
webbnh, line 324: Given that the suffix will presumably be `github.issue.drop`, why are we patching the entry for `github.issue.comment`?

In any case, I think that the reason why this test is passing is because `github.issue.drop` doesn't match any handler...so I don't think this test is functioning as intended (as coded, I think it's redundant with the previous test).

tests/test_main.py
```diff
@@ -289,67 +292,57 @@ def test_initialize_github_error(
...
-            ),
-        ]
+        mock_load_config.return_value = self.mock_config
 
         # Call the function
-        m.listen(self.mock_config)
+        self.old_style_mock_message.topic = "d.d.d.github.issue.comment"
+        self.new_style_mock_message.topic = "d.d.d.github.issue.comment"
+        m.callback(self.old_style_mock_message)
+        m.callback(self.new_style_mock_message)
```
webbnh, line 345: I would put these in two separate code blocks (or otherwise make it clear that these are setups and executions of two separate test scenarios.

tests/test_main.py
ralphbean: Agreed - let's delete this test.

sync2jira/main.py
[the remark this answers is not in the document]
ralphbean: Hm, I'm going to leave this one for another future change.

webbnh (approved):

PR #290, opened 2025-02-04
Improve PR Jira issue matcher

webbnh:
Springboarding from [the discussion here](https://redhat-internal.slack.com/archives/C058M9PBGEM/p1738566886157029?thread_ts=1738345347.500559&cid=C058M9PBGEM), this PR overhauls the PR-to-Jira matcher code:
- Rework the "magic cookie" regex:
  - make it tolerate whitespace better
  - tighten up the match on the Jira reference a little
  - remove redundant syntax
- Replace the build-a-string-and-search-the-whole-thing approach with one which searches each component separately:
  - I don't think this makes the searching more expensive
  - it saves having to instantiate the combined string
  - it means that we cannot hit on matches which inadvertently cross component boundaries
  - we quit searching on the first match instead of gathering them all for checking later
- Pre-compile the regex for efficiency
- Remove a stale/incorrect code comment
- Remove what looks like an unnecessary re-check
- Add type hints to the function signature

This change also adds a bunch of unit test scenarios for the matcher.

### DIFF
--- a/sync2jira/intermediary.py
+++ b/sync2jira/intermediary.py
@@ -17,6 +17,7 @@
 #
 # Authors:  Ralph Bean <rbean@redhat.com>
 import re
+from typing import Optional
 
 
 class Issue(object):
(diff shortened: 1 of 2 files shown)


sync2jira/intermediary.py
```diff
@@ -277,33 +278,37 @@ def map_fixVersion(mapping, issue):
...
+
     for comment in reversed(comments):
-        all_data += f" {comment['body']}"
-    if content:
-        all_data += content
-    if all_data:
-        # Parse to extract the JIRA information. 2 types of matches:
-        # 1 - To match to JIRA issue (i.e. Relates to JIRA: FACTORY-1234)
-        # 2 - To match to upstream issue (i.e. Relates to Issue: !5)
-        match_jira = re.findall(r"Relates to JIRA: ([\w]*-[\d]*)", all_data)
```
webbnh, line 299: @ralphbean, @Zyzyx:  the code only matches case 1, so in my change I've removed this comment.  Is the code _supposed_ to work for case 2??  (If so, I can correct the code, instead....)

sync2jira/intermediary.py
```diff
@@ -277,33 +278,37 @@ def map_fixVersion(mapping, issue):
...
-        if match_jira:
-            for match in match_jira:
-                # Assert that the match was correct
-                if re.match(r"[\w]*-[\d]*", match):
-                    return match
```
webbnh, line 304: @ralphbean, do we know why the code performed this second check?  It seems superfluous and wasteful, so I've removed it...did I miss something??

sync2jira/intermediary.py
```diff
@@ -277,33 +278,37 @@ def map_fixVersion(mapping, issue):
...
-    all_data = " "
+
+    def find_it(input_str: str) -> str:
+        match = JIRA_REFERENCE.search(input_str)
+        return match.group(1) if match else None
+
     for comment in reversed(comments):
-        all_data += f" {comment['body']}"
-    if content:
-        all_data += content
```
webbnh, line 294: @ralphbean, is the `content` (i.e., the PR description) really supposed to be the _last_ resort, such that matches in the comments override it?  It seems like the description (which should be editable after the fact) should rule over the comments...but, I've retained this behavior.  Let me know if I should change it.

webbnh (commented):
Some questions for reviewers....

sync2jira/intermediary.py
webbnh: FWIW, this code seems substantially unchanged since it was originally added in #8...so I guess it _never_ worked for case 2 (the `!` wouldn't match the `-`).

sync2jira/intermediary.py
ralphbean: Yeah, let's retain it for now. It's kind of screwy, but if its what current users expect, let's not upend it without consulting them.

sync2jira/intermediary.py
ralphbean: No, that seems odd. Ok to remove it. It was probably underconfidence in correctly writing the regex in the first place.

ralphbean (approved):

PR #291, opened 2025-02-05
Use simpler "JIRA:" label to match Jira ID in GitHub PRs

hluk:


webbnh:
This looks fine to me:  this new pattern should also match references which use the old pattern, so I don't see any user-compatibility issue.

webbnh:
@ralphbean, is there any documentation of this feature which should be updated?  (If so, where/how?)

hluk:
> Answering my own question, @hluk, you should add [an update to `docs/source/config-file.rst`](https://github.com/release-engineering/Sync2Jira/blob/0c0ed067b51a6ada1dbc4383386f6537c3d8b6c6/docs/source/config-file.rst?plain=1#L115) to this PR.

:+1: Updated.

PR #298, opened 2025-02-28
Guard against none-type inputs to regex matching

ralphbean:
We regularly get tracebacks with the following error:

```
File "intermediary.py", line 302, in find_it match = JIRA_REFERENCE.search(input_str) TypeError: expected string or bytes-like object
```

I'm not sure that this is showing up as a None, but that's the most likely scenario.

### DIFF
--- a/sync2jira/intermediary.py
+++ b/sync2jira/intermediary.py
@@ -299,6 +299,8 @@ def matcher(content: Optional[str], comments: list[dict[str, str]]) -> str:
     """
 
     def find_it(input_str: str) -> str:
+        if input_str is None:
+            return None
         match = JIRA_REFERENCE.search(input_str)
         return match.group(1) if match else None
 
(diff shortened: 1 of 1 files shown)


sync2jira/intermediary.py
```diff
@@ -299,6 +299,8 @@ def matcher(content: Optional[str], comments: list[dict[str, str]]) -> str:
     """
 
     def find_it(input_str: str) -> str:
+        if input_str is None:
+            return None
```
webbnh, line 303: Looks like a good idea, although, if we're gonna do that, we could go one step further:
```suggestion
        if not input_str:
            return None
```
which will skip the search on empty inputs as well as `None` inputs.

webbnh (dismissed):
Looks good, although I have a one-plus suggestion if you like.

sync2jira/intermediary.py
ralphbean: Heh, that `not input_str` was my first impulse, but I went with `input_str is None` thinking "explicit is better than implicit". I'm looking for None here, let me be clear about that.

But, either way. :)

PR #299, opened 2025-02-28
Preserve whitespace to make tracebacks readable

ralphbean:
The traceback emails are hard to read. If we preserve whitespace, they'll hopefully look more natural to the eye.


webbnh (approved):
Looks OK to me.

I might have been inclined to do this with a `<pre>` tag on line 8 instead of using a style, but if the style works, I'm good with that.

PR #301, opened 2025-02-28
Drop the remove-duplicates command

ralphbean:
This isn't a feature that we ever use.

Drop it so that we have less to maintain.


webbnh (approved):

webbnh:
Apparently this went in as part of #302.

PR #308, opened 2025-03-17
fedora-messaging support for pagure

azhuzhu:
Instead of backports with all necessary commits, I reverted the removal of pagure #208, and added some updates


sync2jira/main.py
```diff
@@ -214,12 +232,25 @@ def initialize_issues(config, testing=False, repo_name=None):
...
+        if "issue" not in mapping.get("pagure", {}).get(upstream, {}).get("sync", []):
+            continue
+        if repo_name is not None and upstream != repo_name:
+            continue
+        for issue in u_issue.pagure_issues(upstream, config):
+            try:
+                d_issue.sync_with_jira(issue, config)
+            except Exception as e:
+                log.error(f"Failed on {issue}\nException: {e}")
+                raise
```
webbnh, line 245: As a general practice, you should use `%s`-style format specifiers (instead of f-strings) for log messages, passing the values as arguments, because this defers the formatting operation until after the level filtering:
```suggestion
            except Exception as e:
                log.error("Failed on %r\nException: %s", issue, str(e))
                raise
```

sync2jira/main.py
```diff
@@ -214,12 +232,25 @@ def initialize_issues(config, testing=False, repo_name=None):
...
+    log.info("Done with pagure issue initialization.")
+
     for upstream in mapping.get("github", {}).keys():
         if "issue" not in mapping.get("github", {}).get(upstream, {}).get("sync", []):
             continue
         if repo_name is not None and upstream != repo_name:
             continue
-        # Try and except for GitHub API limit
+        # Try and except for github API limit
```
webbnh, line 253: GitHub should be capitalized.  Ditto line 275, et al.

sync2jira/upstream_issue.py
```diff
@@ -181,6 +181,126 @@ def handle_github_message(body, config, is_pr=False):
...
+
+    :param Dict body: FedMsg Message
+    :param Dict config: Config File
+    :returns: Issue object
+    :rtype: sync2jira.intermediary.Issue
+    """
+    upstream = body["project"]["name"]
+    ns = body["project"].get("namespace") or None
+    if ns:
+        upstream = "{ns}/{upstream}".format(ns=ns, upstream=upstream)
```
webbnh, line 196: The `or None` is pointless:  if the `get()` "fails", it will return `None` anyway (which, will evaluate to `False` and result in the RHS of the `or` being returned which will result in `None`...).

(It's true that the `get()` could return some _other_ False-y value which would then be replaced by `None`, but the next line is going to handle _all_ False-y values, so there is no need to make the replacement.)

sync2jira/upstream_issue.py
```diff
@@ -181,6 +181,126 @@ def handle_github_message(body, config, is_pr=False):
...
+    # Reformat  the assignee value so that it is enclosed within an array
+    # We do this because Github supports multiple assignees, but JIRA doesn't :(
+    # Hopefully in the future it will support multiple assignees, thus enclosing
+    # the assignees in a list prepares for that support
+    for issue in data:
+        issue["assignee"] = [issue["assignee"]]
+
+    issues = (i.Issue.from_pagure(upstream, issue, config) for issue in data)
+    for issue in issues:
+        yield issue
```
webbnh, line 301: Alternatively,
```suggestion
    for issue in data:
        yield i.Issue.from_pagure(upstream, issue, config)
```
which saves creating the extra iterator.

sync2jira/upstream_pr.py
```diff
@@ -20,13 +20,56 @@
...
+    :param Dict body: FedMsg Message
+    :param Dict config: Config File
+    :returns: Issue object
+    :rtype: sync2jira.intermediary.PR
+    """
+    # Extract our upstream name
+    upstream = body["pullrequest"]["project"]["name"]
+    ns = body["pullrequest"]["project"].get("namespace") or None
+    if ns:
+        upstream = "{ns}/{upstream}".format(ns=ns, upstream=upstream)
```
webbnh, line 44: The `or None` is superfluous (see previous comment).

sync2jira/upstream_pr.py
```diff
@@ -55,6 +98,46 @@ def handle_github_message(body, config, suffix):
...
+    data = response.json()["requests"]
+
+    # Reformat Assignee
+    for pr in data:
+        pr["assignee"] = [pr["assignee"]]
+
+    # Build our final list of data and yield
+    prs = (i.PR.from_pagure(upstream, pr, "open", config) for pr in data)
+    for pr in prs:
+        yield pr
```
webbnh, line 138: The extra generator is unnecessary (see previous comment).

webbnh (approved):
I only reviewed [the last six commits](https://github.com/release-engineering/Sync2Jira/pull/308/files/24a48ee1276d035b5ebfca6eba86fef3ef32aba7..f87392b81fecdf32ade98e9be143f58958e794bd#diff-e6b22744f64c1062a0ec04fab35d3ccc574b60e1f4bc34aaaaff7ff5765a50d8R136-R138), and I barely skimmed the tests, but, on that basis, these changes look OK to me.

There are some annoyances, like corrections to the spelling of "GitHub" and the removal of useless line-continuation characters, which this change is reverting, which I've called out below.  And, I've pointed out a few easy fixes for code quality.

But, I wouldn't block the merge for any of those.

azhuzhu:
> I only reviewed [the last six commits](https://github.com/release-engineering/Sync2Jira/pull/308/files/24a48ee1276d035b5ebfca6eba86fef3ef32aba7..f87392b81fecdf32ade98e9be143f58958e794bd#diff-e6b22744f64c1062a0ec04fab35d3ccc574b60e1f4bc34aaaaff7ff5765a50d8R136-R138), and I barely skimmed the tests, but, on that basis, these changes look OK to me.
> 
> There are some annoyances, like corrections to the spelling of "GitHub" and the removal of useless line-continuation characters, which this change is reverting, which I've called out below. And, I've pointed out a few easy fixes for code quality.
> 
> But, I wouldn't block the merge for any of those.

I've updated this PR. Thanks for your review 🚀

sync2jira/intermediary.py
```diff
@@ -116,20 +120,21 @@ def from_pagure(cls, upstream, issue, config):
 
         return cls(
             source=upstream_source,
-            title=issue['title'],
-            url=base + '/%s/issue/%i' % (upstream, issue['id']),
+            title=issue["title"],
+            url=base + "/%s/issue/%i" % (upstream, issue["id"]),
```
webbnh, line 124: Only if you make a change for some other reason...we should be using an f-string (or concatenation) here rather than the Python-2-style string interpolation.  (E.g., see [line 290](https://github.com/release-engineering/Sync2Jira/blob/68a15fa6450f7a444b3952376042786a2342d6b6/sync2jira/intermediary.py#L290).)

webbnh (approved):
Looks great!  Thanks for the extra polishing pass.

(Nice catch on combining the issue loops and your other extra cleanups!)

ralphbean (approved):

sync2jira/intermediary.py
azhuzhu: I filed #310 to make things consistent

PR #311, opened 2025-03-20
Retry Jira failures in case they are due to expired credentials

webbnh:
Anecdotal evidence suggests that _Sync2Jira_ crashes every day at about 15:54 UTC due to some sort of Jira error indicating insufficient credentials.  When the pod restarts, everything works fine again.

Having the logs cycle daily is making it hard to observe other instabilities, so I've implemented a simple stop-gap which I hope will address the credentials problem:  in the code which creates the Jira client object, for each of PRs and Issues, I've added a `try` block around the subroutine calls which receive it; if we encounter a Jira exception, the code will create a fresh client and try the operation again.

If my theory is correct that it is an expiring credentials problem, the retry should succeed and we will avoid crashing the pod (if the retry fails, it will let the exception crash the pod as it does now).  Note that the retry will start the update process from the beginning, which means that any partial updates to the Jira ticket will be made a second time.  Hopefully, the effect from this will be benign, and it will be worth the occasional noise to avoid the crash.

In addition to making this change, I added a few other revisions to my branch:
- I've rearranged the logging in the fedmesg consumer code:
  - it looks like the fedmesg client is logging the start and end of each event, so I removed the debug logging messages from our code; and,
  - I added logging for cases where we do _not_ handle events (the cases that we _do_ handle already issue messages to the log), so this way we will see fewer started- and finished-messages with no disposition between them.
- My IDE was flagging a few things in the modules that I was looking at, so I picked a little lint -- mostly fixing typos in comment text and adding type hints to the code.
- There were a couple of places where I made small code quality tweaks.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -24,7 +24,7 @@ import difflib
 import logging
 import operator
 import re
-from typing import Optional
+from typing import Optional, Union
 
 # 3rd Party Modules
 from jira import JIRAError
@@ -433,7 +433,7 @@ def assign_user(client, issue, downstream, remove_all=False):
     log.warning("Was not able to assign user %s", issue.assignee[0]["fullname"])
 
 
-def change_status(client, downstream, status, issue):
+def change_status(client, downstream, status, issue: Union[Issue, PR]):
     """
     Change status of JIRA issue.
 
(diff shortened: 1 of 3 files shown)


ralphbean (approved):

sync2jira/downstream_issue.py
```diff
@@ -1152,6 +1152,26 @@ def sync_with_jira(issue, config):
         ):
             issue.content = pypandoc.convert_text(issue.content, "jira", format="gfm")
 
+    retry = False
```
ralphbean, line 1155: I think if I were writing this, I would have made `retry = False` into `retries = 0`, and each time you retry, increment the counter instead of setting it to `True`.

Then, the conditional looks like `if retries > 1: raise` or `if retries > 3: raise` to retry more aggressively.

PR #313, opened 2025-03-24
pagure: correct pagure issue topic

azhuzhu:
The actual prefix of Pagure issue topic is "io.pagure.prod.pagure.issue" rather than "org.fedoraproject.prod.pagure.issue"

The one for pull-request is still as-is.


webbnh (approved):
Seems strange, but it's worth a try!

PR #319, opened 2025-04-04
Handle downstream issue which has no description

webbnh:
Apparently, when a Jira issue contains no description, the `description` field is set to `None` in the tool's representation of the downstream issue.  This is awkward when we try to update it, because we apparently expect it to be a string.

This change defends against this case by conditionally using an empty string for the field value if the value is false-y.  (I also moved the relevant assignment from its previous position in the code to the block which includes a description of what is happening.)


ralphbean:
+1 on both counts (the fix, and the line move). Thank you!

ralphbean (approved):

PR #321, opened 2025-04-05
Remove diacritics

webbnh:
For no obvious reason, Jira is returning a `403`/`Access Denied` response when we try to create the downstream issue for [containers/buildah/issues/#6066](https://github.com/containers/buildah/issues/6066).  Since we perform this operation successfully on a routine basis for other upstream issues, and since it fails reliably for this particular issue, I assume that the problem is with the contents of the issue itself.

My best guess is that the name of the upstream reporter, `Jiri Daněk`, is the source of the problem:  the penultimate letter of his surname is not an ASCII or LATIN-1 character -- it's a two-byte or 16-bit character -- and I bet our creaky old Jira service cannot handle it (in two-byte form, the first byte has the high bit set, which might be a problem all by itself).

This PR adds a function which replaces characters that have diacritics with their ASCII counterparts (so `ě` becomes `e`), and applies that function to the issue description before passing it to Jira.  Hopefully, this will allow the creation request to succeed without causing too much damage.  (In Rover, Jiri's surname is spelled with an `e`.)

In addition, while I was editing the `_build_description()` function, I made some small code quality improvements, which I've included in a separate commit:
- refactored the repeated dictionary lookup, caching the result in a local variable
- changed the initialization of the `description` variable to an unconditional assignment with a conditional update, instead of having it instantiated in both branches of the `if` statement
- replaced the Python2 string interpolations with f-strings
- removed the comments from the blocks of imports -- now that we're using `isort`, it will manage these for us, so we don't need the annotations (and, without them, we can properly keep all of the Python standard library packages together in the same block).

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -25,6 +25,7 @@ import logging
 import operator
 import re
 from typing import Any, Optional, Union
+import unicodedata
 
 # 3rd Party Modules
 from jira import JIRAError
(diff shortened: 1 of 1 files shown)


ralphbean (approved):

ralphbean:
After discussing in slack, we don't think this solves a problem that we actually have.

PR #326, opened 2025-04-24
Remove diacritics, take 2

webbnh:
In the course of looking into another issue, I noticed that _Sync2Jira_ has separate code for creating and updating assignees for downstream issues.  PR #325 fixed the name-matching vis a vis diacritic characters when creating issues, but we need a similar tweak when updating existing issues.

This PR enhances the conditional in `_update_assignee()` which compares the up- and downstream assignees so that it performs a second check with diacritics replaced if the first check fails.

While I was in there, I overhauled the code:
- It now uses the boolean flags that it constructs to enable the code to use only a single call to `assign_user()`, rather than having one in each of several branches in the code.
- I replaced the `try` block with a `hasattr()` test.
- Hopefully, the logic is easier to follow, now.

I also changed the function's interface, so that, instead of accepting the whole list of update configuration options, it now takes just a boolean indicating whether to overwrite the assignment or not.  This saves the function from having to construct this value.  This in turn makes things more complicated for the caller, but the calling code was having to do nearly all the work already, so this change saves us from having to do the work _twice_.

Concordant with these changes, I updated the unit tests for `_update_assignee()`.  These seemed to test only a few cases, so I replaced them with a table-driven test which exercises about 20 permutations.

While I was in the testing mindset, I noticed that #325 neglected to include unit test coverage for the code that it added, so there is a second commit in this PR which adds two more unit tests toward that end.  While I was at it, I fixed a "bug" where we called a mock method instead of calling the corresponding assertion.  (This commit also improves on the function header comments for some of the related tests, which were pretty much just copies of each other.)

Finally, my IDE just updated to a new version, and now it is pickier than ever, especially with respect to code comments.  So, I've included a third commit which addresses those complaints.  I recommend reviewing each of the three commits individually -- the last one is almost entirely non-functional changes, and the actual changes to the code are removal of unused variables and a cosmetic addition to a `return` statement.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -691,9 +691,14 @@ def _update_jira_issue(existing, issue, client, config):
         _update_fixVersion(updates, existing, issue, client)
 
     # Only synchronize assignee for listings that op-in
-    if any("assignee" in item for item in updates):
-        log.info("Looking for new assignee(s)")
-        _update_assignee(client, existing, issue, updates)
+    for item in updates:
+        if isinstance(item, dict):
+            if assignee := item.get("assignee"):
+                log.info("Looking for new assignee(s)")
+                _update_assignee(
+                    client, existing, issue, assignee.get("overwrite", False)
+                )
+                break
 
     # Only synchronize descriptions for listings that op-in
     if "description" in updates:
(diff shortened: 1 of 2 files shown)


ralphbean (approved):
Thanks for this @webbnh!

PR #332, opened 2025-05-06
Fix upstream assignee of `None` and remove a superfluous logging argument

webbnh:
This PR addresses a couple of faults in #327.

First, one of the logging calls tweaked in that PR ended up with an extra argument parameter, and this causes the call to fail.  This PR removes the superfluous argument.

Second, it turns out that an upstream issue can have an assignee with a `"fullname"` key value of `None`, which triggers an exception when we try to treat it as a string.  This PR addresses this in two ways:  it detects the situation and avoids referencing the value, considering the issue to be "unassigned" (rather than "assigned to nothing"); but, also, the function which was incurring the exception now guards against the condition, as well.

In addition to those changes, this PR adds a case to the applicable unit test to generate this condition, to ensure that it is handled appropriately.  (And, I reworked the test case slightly, since the addition of one more wrinkle made the existing code kind of hokey.)

Finally, I added a unit test for `remove_diacritics()`.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -437,7 +437,6 @@ def assign_user(
         downstream.key,
         [a["fullname"] for a in issue.assignee],
         issue.url,
-        [a["fullname"] for a in issue.assignee],
     )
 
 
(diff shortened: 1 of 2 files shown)


ralphbean (approved):
Awesome, thank you Webb!

PR #338, opened 2025-05-14
Use GitHub issue type (if present) for the Jira issue type

### LINKED ISSUE #330: Add ability to sync JIRA issue type with GH issue type
From this conversation in slack: https://redhat-internal.slack.com/archives/C058M9PBGEM/p1737495007203029

Currently, the only way for sync2jira to give jira a type is from GH labels. GH now has a a a type field that we'd like to be able to sync to jira on.

webbnh:
Currently, when _Sync2Jira_  creates a downstream issue, it determines the issue type using the following order of preference:
 - an issue type from the mapping in the configuration file (if present), selected based on the upstream "tags" (labels);
 - the default issue type configured for the project (if any);
 - `Story` if the issue title contains "RFE";
 - otherwise, `Bug`.

However, GitHub was recently enhanced to allow [Organizations to define "types"](https://docs.github.com/en/issues/tracking-your-work-with-issues/configuring-issues/managing-issue-types-in-an-organization) for upstream issues, providing a few types for use by default (`Task`, `Bug`, and `Feature`).

This PR changes the code such that, when _Sync2Jira_ creates a new Jira issue, the GH issue type, if set, will be the first preference.

_**Note:** the GH issue type **must** match one of the issue types provided by the downstream project's schema or the creation will fail and the issue will not be sync'd._

**Question for reviewers:**  Should the settings in the configuration file precede the upstream type in the order of preference?  If this PR is merged as is, there will be no way to handle the situation if the upstream Organization uses types which aren't available in the downstream project.  (Fortunately, the default GitHub types look like they all match Jira types, although I expect that `Feature` has very different semantics for most downstream projects.)

--------

In addition, this PR contains the following point changes/fixes:
- The code which adds a comment to the downstream issue noting the fact that there was more than one plausible issue type was throwing an exception owing to what amounts to a syntax error resulting from a botched edit (of mine...  😊  #254); the second commit fixes this, and the first includes a new unit test case for it.
- One of my changes from earlier this year (#282) identified a gap in the code where we were ignoring a certain type of FedMsg and added a log message for it.  I've since decided that we _should_ (explicitly) ignore this case, so I replaced the `FIXME` comment with an explanation and downgraded the log message from `INFO` to `DEBUG` level.
- In the common upstream issue processing path, we have a conditional which allows control to escape if the "issue" is actually a PR (and if the message is not indicating that the PR is now closed); I made the test slightly more robust (it will now properly handle the case of the key being present with an empty value).  
- I repositioned the log message which reports the inability to assign the downstream issue and made it conditional on there actually being someone to attempt to assign it to.  This should reduce the frequency of spurious occurrences in the log; therefore, I upgraded it from `INFO` to `WARNING` level.
- I reworked the body of `_get_preferred_issue_types()`, hopefully making it a little simpler and tighter.  I moved the inline code comments to the function docstring, and I added an exhaustive unit test for it.
- I modified the intermediary issue object to handle the GH issue type as an additional field.  I added the base case to an existing unit test as well as adding an additional unit test for the various possible cases.
- I refactored some duplicated code in the intermediary objects, moving it to a common function.
- I picked a bunch of lint that my IDE was flagging.

--------
Fixes #330.
Fixes DPROD-769.
cc: @gitdallas

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -484,9 +484,19 @@ def change_status(client, downstream, status, issue: Union[Issue, PR]):
 
 def _get_preferred_issue_types(config, issue):
     """
-    Consider the configuration file and issue data to
-    figure out the right issue type to file. Used when
-    creating a new issue in Jira.
+    Determine the appropriate issue type to specify when creating the
+    downstream (Jira) issue.  In order of preference:
+     - the upstream issue type (if any)
+     - the issue type(s) from the mapping in the configuration file (if
+        present), selected based on the upstream "tags" (labels)
+     - the default issue type configured for the project (if any)
+     - "Story" if the issue title contains "RFE"
+     - otherwise, "Bug".
+
+    In all cases, a list of one item is returned, except when the upstream
+    issue has multiple tags which match multiple entries in the configured
+    mapping, in which case multiple entries are returned, sorted in ascending
+    lexographical order.
 
     :param Dict config: Config dict
     :param sync2jira.intermediary.Issue issue: Issue object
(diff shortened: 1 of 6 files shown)


gitdallas:
i wonder if the labels=>type should have precedence as it requires the user to set up a config to support them. if they don't want labels=>type they would leave out the config, i believe.

or possibly allow a configuration to decide

webbnh:
Thanks for the feedback, @gitdallas!  I concur, so I've change the code as follows:  if there is a control in the configuration file which is applicable (either a label-based mapping or a specified issue type), then we'll use that for the downstream issue type; otherwise, we'll try to use the upstream type if there is one (and, if we fail, then the downstream issue won't get created); otherwise, we'll use the existing fallbacks of `Story` (if the title contains `"RFE"`) or `Bug`.

That sounds least likely to disrupt existing usage, while still allowing the new GitHub feature to have the desired effect; and, if the upstream causes problems (which cannot be fixed by RH), then we can address them using the config file.

Thanks!

ralphbean (approved):

PR #339, opened 2025-05-20
Validate return from GitHub client `get_user()`

webbnh:
Since improving the logging, we're now seeing some messages like
```
Unable to assign RUN-3076 from upstream assignees [None] in https://github.com/containers/buildah/issues/6152
```
despite my attempts to suppress this message in cases where the issue is "unassignable".

Apparently (although, I have scant direct proof for this...), when the upstream user's profile doesn't include a publicly-available human-readable name (as is the case for the current assignee in https://github.com/containers/buildah/issues/6152), the GitHub client's `get_user()` function returns a value with `"None"` (not a value of `None`!) as the name.

So, this PR catches and ignores that value, which, in the case of the log message above, should result in an _empty_ list of assignees, which will in turn result in the message being suppressed.

### DIFF
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -327,7 +327,7 @@ def reformat_github_common(item, github_client):
     # Search for the user
     reporter = github_client.get_user(item["user"]["login"])
     # Update the reporter field in the message (to match Pagure format)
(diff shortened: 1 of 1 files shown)


ralphbean (approved):
Oh, github.

PR #349, opened 2025-06-26
Add latest pandoc to the sync-page container image, too

webbnh:
This is a follow-on to #342, which should have included this change as well.

This PR makes the same change to the `sync-page` container as #342 made to the `sync2jira` container.

### DIFF
--- a/Dockerfile.sync-page
+++ b/Dockerfile.sync-page
@@ -45,10 +45,13 @@ RUN mkdir -p /usr/local/src/sync2jira
 COPY . /usr/local/src/sync2jira
 
 # Install deps
-RUN pip install -r /usr/local/src/sync2jira/requirements.txt
+RUN pip3 install -r /usr/local/src/sync2jira/requirements.txt
+
+# Grab the latest pandoc binary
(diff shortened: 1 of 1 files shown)


webbnh:
@ralphbean, rather than maintaining two very-nearly-identical Dockerfiles, what do you think of building the `sync-page` image from the `sync2jira` one?  I think then the contents of `Dockerfile.sync-page` would look basically like this:
```
FROM <sync2jira-image-ref>

LABEL \
    name="sync2jira-sync-page" \
    org.opencontainers.image.name="sync2jira-sync-page"

CMD ["python3", "/usr/local/src/sync2jira/sync-page/event-handler.py"]
```
And that's it.  (We _might_ have to specify **all** the `LABEL` vales...I'm not positive whether the non-conflicting ones in the base image are retained or not, but that's small potatoes.)  (And, it'll probably build _a lot_ faster....)

The tricky part is figuring out how to supply the reference to the `sync2jira` image to use as the base image, but I'm hoping that's not too hard.

Thoughts?

ralphbean (approved):

ralphbean:
We could also probably consolidate down to just _one_ image and simply override the entry point in the kube deployment config.

ralphbean:
But, if for some reason that doesn't work out - yeah, building one image from the other works!

PR #363, opened 2025-08-12
Enable retry of Jira requests for PR syncing

webbnh:
The logs recently held the following sequence:
```
[2025-08-06 09:03:30,377] INFO: Consuming message from topic org.fedoraproject.prod.github.pull_request (message id d68e7253-47e5-48c6-bf42-27f8672f00d5)
[2025-08-06 09:03:30,481] INFO: [PR] Considering upstream https://github.com/quipucords/quipucords-ui/pull/655, [quipucords/quipucords-ui] chore(release): 2.0.1
[2025-08-06 09:03:36,895] WARNING: No JIRA issue exists for PR: [quipucords/quipucords-ui] chore(release): 2.0.1. Query: Key = DISCOVERY-1054
[2025-08-06 09:03:36,896] INFO: Successfully consumed message from topic org.fedoraproject.prod.github.pull_request (message id d68e7253-47e5-48c6-bf42-27f8672f00d5)
```
However, contrary to the warning, the Jira issue, `DISCOVERY-1054`, _did_ exist at the time.  Instead, the request for it failed due (apparently) to one of our frequent, mysterious, transient Jira server errors.  And, the code involved caught and "handled" the error, rather than letting it propagate back to the logic which would have re-tried the request.

This PR removes the `try` block and the inappropriate error message.  Without it, the failure will be handled in the caller and retried and/or the actual Jira error will be logged.  The change also reworks the handling of the response, placing the success case first and splitting the failure cases.  It also downgrades the cases which are driven by the contents of the PR (as opposed to problems with the operation of Jira or the tool itself) from warnings to informational messages.  Finally, it adds an escape which skips the Jira request if we don't have an issue to request.

This pull request also includes a new unit test (so that rather than decreasing the coverage by 0.11% it _increases_ by 0.09% 🤪).  And, in the course of writing that, I stumbled upon a typo in a mock reference which (literally) does nothing, but, when you correct the typo, the test _fails_...so I removed the mock (and the incorrect and unneeded assertion).

### DIFF
--- a/sync2jira/downstream_pr.py
+++ b/sync2jira/downstream_pr.py
@@ -213,24 +213,24 @@ def update_jira(client, config, pr):
 
     # Find our JIRA issue if one exists
     if isinstance(pr, Issue):
+        # Instances of the `PR` type already have an initialized `jira_key`
+        # attribute; since what we have here is actually an `Issue`, we need to
+        # add it.
         pr.jira_key = matcher(pr.content, pr.comments)
 
-    query = f"Key = {pr.jira_key}"
-    try:
-        response: ResultList[JIRAIssue] = client.search_issues(query)
-        # Throw error and return if nothing could be found
-        if len(response) == 0 or len(response) > 1:
-            log.warning(f"No JIRA issue could be found for {pr.title}")
-            return
-    except JIRAError:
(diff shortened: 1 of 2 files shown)


ralphbean:
Nice, thanks Webb!

PR #379, opened 2025-10-16
Switch Jira client function for user lookup

webbnh:
The Jira client function which we currently use for looking up users, `search_assignable_users_for_issues()`, seems to have a few problems.  It requires that either the `username` parameter or the `query` parameter be provided, and the `username` parameter has been deprecated in favor of the `query` parameter.  Now, while Jira Cloud appears to still support the `username` parameter, at the point when we call this function, we don't _have_ a Jira username, so that parameter isn't usable for us; on the other hand, while the client accepts the `query` parameter, the Jira Datacenter server seems to ignore it, and it simply returns _all_ the possible users.  In PR #377, I added code to defend against this, but, in some cases (i.e., for some projects), there are too many assignable users for the server to return in a single response (it seems to be limited to 100), so it returns only a partial list.  The API offers facilities for "paging" the list, but I was unable to get them to work (no matter what I do, only the first page seems to be returned).

This PR replaces the use of `search_assignable_users_for_issues()` with `search_users()`, which is willing to accept an email address as a selector and seems to return the correct unique value in the cases that were causing problems.  So, this is a "one-liner" change, except for all of the mocking in the unit tests....

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -377,11 +377,8 @@ def match_user(
     """
 
     for email in emails:
-        # Get a list from Jira of users that match the supplied email address
-        # and that are suitable for assignment to this issue.
-        users = client.search_assignable_users_for_issues(
-            query=email, issueKey=downstream.key
-        )
+        # Get a list from Jira of users that match the supplied email address.
+        users = client.search_users(user=email)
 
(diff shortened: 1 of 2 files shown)


ralphbean:
Thanks so much for working on this!

PR #386, opened 2025-10-24
Replace old "master" references with "main"

ralphbean:


webbnh (commented):
Looks good, but what about `docs/source/conf.py` (specifically [line 7](https://github.com/release-engineering/Sync2Jira/blob/8ea45355c970fb264495d83e883f66771728f8dd/docs/source/conf.py#L7), but there are a couple of other references to "master" (e.g., `master_doc`) which are ripe for expunging)?

PR #387, opened 2025-10-24
Touch ups for #384 and #385

webbnh:
Looking into something else, I noticed a couple of nits introduced by my previous PRs.

The doc addition in #384 renders with an undesired line break and incorrect indentation.  This PR corrects the formatting in the documentation source to wrap the lines properly, so that the text will (hopefully) "flow" properly when rendered.  It also does the same for a number of other long source lines (and, it adds a missing period and removes an extra space).

The change in #385 left `_update_on_close()` with an unused parameter.  This PR removes that parameter from the function signature and from each of the callers, and it removes the setup for it from the unit tests.

### DIFF
--- a/.tekton/build-pipeline.yaml
+++ b/.tekton/build-pipeline.yaml
@@ -107,7 +107,7 @@ spec:
       - name: name
         value: init
       - name: bundle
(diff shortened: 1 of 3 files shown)


ralphbean (approved):
LGTM. Thanks @webbnh!

PR #389, opened 2025-11-07
update code for connecting with snowflake account using private key.

Bala-Sakabattula:
made changes to connect with snowflake service account using private key.


webbnh (commented):
@Bala-Sakabattula, I've flagged a few gaps in the tests for your consideration.

webbnh (approved):
/lgtm

Thanks, Bala!

PR #392, opened 2025-11-14
Fail fast if JIRA authentication isn't working

ralphbean:
Otherwise, this may fail *silently* and just return an empty list [] for a .search_issues query, or it may fail loudly later when we try to update an issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ fedmsg.d/sync2jira.py
 *.pyc
 __pycache__
 .vscode/settings.json
+.worktrees/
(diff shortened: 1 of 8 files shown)


webbnh (approved):
Looks good to me.

However, I do have one thought:  do we really care (i.e., is there a good reason for the tests to assert) that `get_jira_client()` calls `session()` (and calls it exactly once)?  As long as the function returns the expected response or raises the expected exception, isn't that good enough?

I don't want the tests to over-constrain the implementation, but I also suspect that we have better things to do than to hold up this PR over that detail.  😵‍💫

PR #398, opened 2025-11-25
Add cache for linking downstream to upstream issues

webbnh:
The current approach of using Snowflake to find downstream issues which have been linked to a given upstream issue works fine except for during the interval between when `sync2jira` creates a new downstream issue and when that update is propagated to Snowflake.  If an update is made to the upstream issue during that interval, then the Snowflake lookup will fail, and `sync2jira` will create _another_ downstream issue.  This will repeat with each upstream update until Snowflake is updated.  After that point, the next upstream update will sync with one of the downstream issues, but I'm not sure that we can predict which one.  This PR attempts to address both of these issues.

The fix for the multiple creations is nearly trivial.  This PR adds a Python `dict` which is used as a cache in which to look up the downstream issue key using the upstream issue URL.  When we look up an issue, we first check the cache; if we hit in the cache, then we skip the Snowflake query, etc. and proceed.  If the cache lookup misses, then we call Snowflake and proceed normally, except that, once we've selected the downstream issue, we add it to the cache.  Likewise, if we miss completely and create a new issue, we add that result to the cache, as well.

This means that, as long as the tool runs continuously, we'll accumulate a local mapping of issues.  This should be more performant than calling Snowflake and Jira, which is a plus, and it will also close the gap created by the delay between updating Jira and updating Snowflake.  However, it poses a small concern around accumulating memory.  To address this, the cache is implemented using a subclass of `dict` which limits the number of entries (currently to 1000, which is probably many months of `sync2jira` runtime...longer than the typical interval between redeployments for configuration updates).  If the dictionary reaches the limit, it will delete entries in FIFO order as new entries are added.  This should pose no problem as long as the time required to fill the dictionary is larger than the update interval between Jira and Snowflake (which it should be, with no problems), because a cache miss will simply trigger a Snowflake query which should, at that point, return successfully.

This PR also attempts to fix the non-determinism in the selection process when multiple downstream issues are found for a given upstream issue.  With this change, the potential issues should be sorted by last-update time, and the most recently updated issue should be selected.  In the previous code, the sorting was taking place only in an unusual situation, and it was sorting the list backwards and taking the _least_-recently updated issue.

In addition to those changes, this PR also contains several other commits:
- In several (but not all) places in the code, when a configuration omits the `"issues_updates"` option, we were using an empty `dict` as a default value; this PR changes that to be an empty list, as it should be (an empty `dict` did not pose a problem, because in many contexts a `dict` behaves like a `list` of its keys, but this change corrects the type).
- Add a change to try to set the logging level for the Snowflake module's logger to `WARNING`, to quiet down a frequent `INFO` message which isn't useful for our purposes.
- Refactor the code in `_find_comment_in_jira()`, `check_comments_for_duplicate()`, and `_matching_jira_issue_query()` to improve the quality.

This PR also includes a number of improvements for the unit tests for `_find_comment_in_jira()` and some tweaks to the unit test for `_matching_jira_issue_query()` to adapt to the refactor of `check_comments_for_duplicates()` and the addition of the cache.

It looks like the unit testing for `_matching_jira_issue_query()` is pretty light, but I don't have time right now to address that.  I'll be on PTO for the rest of this week, so I'll toss the ball to @Bala-Sakabattula and @ralphbean, if you want to try to get this deployed before I get back.

Happy Thanksgiving!  🦃

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -245,18 +245,21 @@ def _matching_jira_issue_query(client, issue, config):
     :rtype: List
     """
 
-    # If there's an entry for the issue in our cache, return it.
+    # If there's an entry for the issue in our cache, fetch the issue key from it.
     if result := jira_cache.get(issue.url):
-        return result
+        issue_keys = [result]
+    else:
+        # Search for Jira issues with a "remote link" to the issue.url;
+        # if we find none, return an empty list.
+        results = execute_snowflake_query(issue)
+        if not results:
+            return []
 
-    # Search for Jira issues with a "remote link" to the issue.url;
-    # if we find none, return an empty list.
-    results = execute_snowflake_query(issue)
-    if not results:
-        return []
+        # From the results returned by Snowflake, make an iterable of the
+        # issues' keys.
+        issue_keys = (row[0] for row in results)
 
-    # Fetch the Jira issue objects using the keys returned from Snowflake
-    issue_keys = (row[0] for row in results)
+    # Fetch the Jira issue objects using the key list.
     jql = f"key in ({','.join(issue_keys)})"
     results = client.search_issues(jql)
 
@@ -303,7 +306,7 @@ def _matching_jira_issue_query(client, issue, config):
             results = [results[0]]  # A list of one item
 
     # Cache the result for next time and return it.
-    jira_cache[issue.url] = results
+    jira_cache[issue.url] = results[0].key
     return results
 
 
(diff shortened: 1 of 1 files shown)


sync2jira/downstream_issue.py
Bala-Sakabattula: I have a small doubt. When _update_jira_issue() updates a JIRA issue, the issue's updated with changes in JIRA, but the cache still holds the old issue object with stale data.
If multiple JIRA issues link to the same GitHub URL, and one issue gets updated externally, the cache may continue to return the previously cached issue instead of the now-most-recent one. This could cause updates to go to the wrong issue.

PR #399, opened 2025-11-30
Custom Field Mapping

Bala-Sakabattula:
I updated the code to retrieve the IDs using the field name and implemented a cache to store both field names and their corresponding IDs. The field ID is now fetched directly from the cache, and if the value is already an ID, it remains unchanged.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -68,6 +68,7 @@ FROM
 
 
 GH_URL_PATTERN = re.compile(r"https://github\.com/[^/]+/[^/]+/(issues|pull)/\d+")
+field_name_cache = {}
 
 
 class UrlCache(dict):
@@ -152,6 +153,26 @@ def execute_snowflake_query(issue):
     return results
 
 
+def _build_field_name_cache(client):
+    """Build the field name cache for the given JIRA client."""
+    try:
+        # clearing the cache to remove any stale entries
+        field_name_cache.clear()
+
+        # adding the standard fields to the cache
+        for field in ("priority", "assignee", "summary", "description"):
+            field_name_cache[field] = field
+
+        # fetching the custom fields from the JIRA client
+        all_fields = client.fields()
+        name_map = {field["name"]: field["id"] for field in all_fields}
+
+        # updating the cache with the custom fields
+        field_name_cache.update(name_map)
+    except Exception as e:
+        log.error(f"Error building field name cache: {e}")
+
+
 def _get_field_id_by_name(client, field_name):
     """
     Convert a human-readable custom field name to its JIRA field ID.
@@ -161,33 +182,15 @@ def _get_field_id_by_name(client, field_name):
     :returns: Field ID (e.g., "customfield_12310243") or None if not found
     :rtype: Optional[str]
     """
-    # Cache field mappings per client to avoid repeated API calls
-    cache_key = id(client)
-    if cache_key not in jira_cache:
-        jira_cache[cache_key] = {}
-
     # Check cache first
-    if field_name in jira_cache[cache_key]:
-        return jira_cache[cache_key][field_name]
+    if field_ID := field_name_cache.get(field_name):
+        return field_ID
+    # If not in cache, build the cache
+    _build_field_name_cache(client)
 
-    # Fetch all fields from JIRA
-    try:
-        all_fields = client.fields()
-        name_map = {field["name"]: field["id"] for field in all_fields}
-
-        # Cache the entire mapping
-        jira_cache[cache_key] = name_map
-
-        # Return the field ID if found
-        field_id = name_map.get(field_name)
-        if field_id:
-            log.debug(f"Resolved field name '{field_name}' to ID '{field_id}'")
-        else:
-            log.warning(f"Field name '{field_name}' not found in JIRA fields")
-        return field_id
-    except Exception as e:
-        log.error(f"Error fetching JIRA fields: {e}")
-        return None
+    if field_ID := field_name_cache.get(field_name):
+        return field_ID
+    return None
 
 
 def _resolve_field_identifier(client, field_identifier):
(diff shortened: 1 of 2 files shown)


sync2jira/downstream_issue.py
```diff
@@ -152,6 +152,68 @@ def execute_snowflake_query(issue):
...
+    :returns: Field ID or None if not found
+    :rtype: Optional[str]
+    """
+    # If it's already an ID (starts with 'customfield_' or is a standard field), return as-is
+    if field_identifier.startswith("customfield_") or field_identifier in [
+        "priority",
+        "assignee",
+        "summary",
+        "description",
+    ]:
```
webbnh: A stylistic nit:  prefer immutable objects to mutable ones, especially for values which are constant -- use a `tuple` instead of a `list` here:
```suggestion
    if field_identifier.startswith("customfield_") or field_identifier in (
        "priority",
        "assignee",
        "summary",
        "description",
    ):
```

sync2jira/downstream_issue.py
```diff
@@ -152,6 +152,68 @@ def execute_snowflake_query(issue):
...
+    :rtype: Optional[str]
+    """
+    # Cache field mappings per client to avoid repeated API calls
+    cache_key = id(client)
+    if cache_key not in jira_cache:
+        jira_cache[cache_key] = {}
+
+    # Check cache first
+    if field_name in jira_cache[cache_key]:
+        return jira_cache[cache_key][field_name]
```
webbnh: This idiom is unsatisfying because it does the `dict` lookup _twice_ (on what is supposed to be the _fast_ path) -- once for the `in` operation and again for the `[]` reference.


I'd recommend something like this:
```suggestion
    field_ID = jira_cache[cache_key].get(field_name)
    if field_ID:
        return field_ID
```

And, I would recommend using the "walrus" operator for this:
```suggestion
    if field_ID := jira_cache[cache_key].get(field_name)
        return field_ID
```

sync2jira/downstream_issue.py
```diff
@@ -152,6 +152,68 @@ def execute_snowflake_query(issue):
...
+
+    :param jira.client.JIRA client: JIRA client
+    :param str field_name: Human-readable field name (e.g., "Story Points", "Epic Link")
+    :returns: Field ID (e.g., "customfield_12310243") or None if not found
+    :rtype: Optional[str]
+    """
+    # Cache field mappings per client to avoid repeated API calls
+    cache_key = id(client)
+    if cache_key not in jira_cache:
+        jira_cache[cache_key] = {}
```
webbnh: I'm hoping that you're going to just delete this code; however, if you decide to keep it, then there's no point to initializing `jira_cache[cache_key]` to `{}`:  instead, you want the code which calls `client.fields()`...because you _know_ that the lookup in the next block is going to fail.

sync2jira/downstream_issue.py
```diff
@@ -1071,7 +1140,20 @@ def _update_github_project_fields(
...
+                    continue
+                # Resolve the field identifier to an ID
+                jirafieldname = _resolve_field_identifier(client, field_identifier)
+                if not jirafieldname:
+                    log.error(
+                        f"Could not resolve custom field '{field_identifier}' to an ID, skipping"
+                    )
+                    continue
+
                 log.info(f"Jira issue story point field name is:  '{jirafieldname}'")
```
webbnh: While you're in here, consider changing this log message from `INFO`-level to `DEBUG`.

tests/test_downstream_issue.py
```diff
@@ -457,6 +457,7 @@ def common_test_create_jira_issue(
             {"name": "Epic Link", "id": "customfield_1"},
             {"name": "QA Contact", "id": "customfield_2"},
             {"name": "EXD-Service", "id": "customfield_3"},
+            {"name": "somecustumfield", "id": "customfield_4"},
         ]
```
webbnh: Should we have any of the "special case" fields (like `"summary"`) in this list?

(Also, is the spelling of `somecustumfield` deliberate, or should it be `somecustomfield`?  If you change it, you should probably change the value at line 472 to match.  😇)

Ditto for the list at line 595 (and at line 612).

tests/test_downstream_issue.py
```diff
@@ -1831,6 +1838,173 @@ def test_remove_diacritics(self):
...
+
+    def test_resolve_field_identifier(self):
+        """Test _resolve_field_identifier function"""
+        # Clear cache first
+        d.jira_cache.clear()
+
+        mock_client = MagicMock()
+        mock_client.fields.return_value = [
+            {"name": "Story Points", "id": "customfield_12345"},
+        ]
```
webbnh: The code in `_resolve_field_identifier()` doesn't reference `d.jira_cache`, and its subroutine, `_get_field_id_by_name()` already has its own unit test, so it doesn't need to be tested here.  That is, instead of maintaining the cache here, we should be mocking `_get_field_id_by_name()` and setting its return value instead of using the client mock.

sync2jira/downstream_issue.py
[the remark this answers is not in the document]
Bala-Sakabattula: I am using the update functionality if the client.fields returns the special fields they will get updated.

``` python
global field_name_cache
# clearing the cache to remove any stale entries
field_name_cache.clear()

# adding the standard fields to the cache
for field in ("priority", "assignee", "summary", "description"):
    field_name_cache[field] = field

# fetching the custom fields from the JIRA client
all_fields = client.fields()
name_map = {field["name"]: field["id"] for field in all_fields}

# updating the cache with the custom fields
field_name_cache.update(name_map)
```

tests/test_downstream_issue.py
Bala-Sakabattula: Thanks for pointing that out! In the setup function of the test class, the mock issue is configured to use somecustomfield, so I kept the same field name in the related tests to stay consistent with the existing mock data.

``` python
self.mock_issue.downstream = {
            "project": "mock_project",
            "custom_fields": {"somecustumfield": "somecustumvalue"},
            "qa-contact": "dummy@dummy.com",
            "epic-link": "DUMMY-1234",
            "EXD-Service": {"guild": "EXD-Project", "value": "EXD-Value"},
            "issue_updates": [
                "comments",
                {"tags": {"overwrite": False}},
                {"fixVersion": {"overwrite": False}},
                {"assignee": {"overwrite": True}},
                "description",
                "title",
                {"transition": "CUSTOM TRANSITION"},
                {"on_close": {"apply_labels": ["closed-upstream"]}},
            ],
            "owner": "mock_owner",
        } 
```

ralphbean (approved):

webbnh (commented):
It's good that you've moved forward with this during my absence, but I have a couple of comments in case you want to open another PR.  However, I would like to continue the discussion on whether to create an issue or abort if we run into problems with referenced custom fields.

PR #400, opened 2025-12-02
Jira cache entries are strings, not lists

webbnh:
In #398, we ended up confused about what the entries in the `jira_cache` looked like.  This causes the code to fail with a `TypeError` (and so we avoid creating duplicate issues by dying instead!).  This PR reconciles the situation:  they are strings, not lists (of one string).

This PR corrects `_create_jira_issue()` to match the behavior of `_matching_jira_issue_query()`.

And, since we never are interested in a retrieving more than one issue, this PR also in-lines `_matching_jira_issue_query()` into `_get_existing_jira_issue()` to remove the extra, unneeded layer of call, changing the combined result to return a string (or `None`) instead of a list.

This PR also tweaks the one unit test to allow it to work.  I still need to add proper unit testing for the function which is now called `_get_existing_jira_issue()`, but the function is too complicated to test easily, so it needs to be refactored first, which I'll do in a separate PR.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -234,26 +234,26 @@ def get_jira_client(issue, config):
     return client
 
 
-def _matching_jira_issue_query(client, issue, config):
+def _get_existing_jira_issue(client, issue, config):
     """
-    API calls that find matching JIRA tickets if any are present.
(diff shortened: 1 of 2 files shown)


Bala-Sakabattula (approved):
LGTM!

PR #404, opened 2026-01-06
Prevent issue creation when custom field mapping fails

Bala-Sakabattula:
This PR adds a validation check to ensure that issue creation is stopped when custom field mapping fails.

### DIFF
--- a/tests/test_downstream_issue.py
+++ b/tests/test_downstream_issue.py
@@ -1954,18 +1954,12 @@ class TestDownstreamIssue(unittest.TestCase):
     ):
         """Test _update_github_project_fields raises ValueError when storypoints field cannot be resolved"""
         github_project_fields = {"storypoints": {"gh_field": "Estimate"}}
-        original_storypoints = self.mock_config["sync2jira"]["default_jira_fields"][
-            "storypoints"
-        ]
-        self.mock_config["sync2jira"]["default_jira_fields"][
-            "storypoints"
-        ] = "Story Points"
-
+        jira_fields = self.mock_config["sync2jira"]["default_jira_fields"]
+        original_storypoints = jira_fields["storypoints"]
+        jira_fields["storypoints"] = "Story Points"
         # Set up mock - storypoints field identifier doesn't exist in JIRA
-        mock_client.fields.return_value = [
-            {"name": "Priority", "id": "priority"},
-            # Note: The field from default_jira_fields.storypoints doesn't exist
-        ]
+        # Note: The field from default_jira_fields.storypoints doesn't exist
+        mock_client.fields.return_value = [{"name": "Priority", "id": "priority"}]
 
         # Clear cache to force fresh lookup
         d.field_name_cache.clear()
@@ -1982,7 +1976,7 @@ class TestDownstreamIssue(unittest.TestCase):
 
         # Assert error message contains diagnostic information
         error_msg = str(context.exception)
-        self.assertIn("story points", error_msg.lower() or error_msg)
+        self.assertIn("story points", error_msg.lower())
         self.assertIn("Could not resolve", error_msg)
         # Issue should not be updated
         self.mock_downstream.update.assert_not_called()


sync2jira/downstream_issue.py
```diff
@@ -155,22 +155,24 @@ def execute_snowflake_query(issue):
...
 
-        # adding the standard fields to the cache
-        for field in ("priority", "assignee", "summary", "description"):
-            field_name_cache[field] = field
+    # clearing the cache to remove any stale entries
+    field_name_cache.clear()
+
+    # adding the standard fields to the cache
+    for field in ("priority", "assignee", "summary", "description"):
+        field_name_cache[field] = field
```
webbnh: Alternatively,
```suggestion
    # Reset the cache to just the standard fields
    field_name_cache = {f: f for f in ("priority", "assignee", "summary", "description")}
```

sync2jira/downstream_issue.py
```diff
@@ -155,22 +155,24 @@ def execute_snowflake_query(issue):
...
+    # fetching the custom fields from the JIRA client
+    try:
         all_fields = client.fields()
         name_map = {field["name"]: field["id"] for field in all_fields}
 
         # updating the cache with the custom fields
         field_name_cache.update(name_map)
     except Exception as e:
         log.error(f"Error building field name cache: {e}")
+        raise
```
webbnh: I recommend tightening the `try` block.  And, did you want to avoid the extra `dict` instantiation?
```suggestion
    # fetching the custom fields from the JIRA client
    try:
        all_fields = client.fields()
    except Exception as e:
        log.error(f"Error building field name cache: {e}")
        raise

    # updating the cache with the custom fields
    for field in all_fields:
        field_name_cache[field["name"]] = field["id"]
```

tests/test_downstream_issue.py
```diff
@@ -1910,11 +1910,13 @@ def test_get_field_id_by_name_exception(self):
...
         # Call function
-        result = d._get_field_id_by_name(mock_client, "Story Points")
+        with self.assertRaises(Exception) as context:
+            result = d._get_field_id_by_name(mock_client, "Story Points")
 
         # Assert
-        self.assertIsNone(result)
+        self.assertEqual(str(context.exception), "Connection error")
         mock_client.fields.assert_called_once()
+        self.assertEqual(len(d.field_name_cache), 4)
```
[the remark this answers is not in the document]
Bala-Sakabattula: Okay I will change it to match the standard fields only and remove the length

sync2jira/downstream_issue.py
```diff
@@ -732,8 +732,10 @@ def _create_jira_issue(client, issue, config):
         # If key is a field name, resolve it to an ID
         field_id = _resolve_field_identifier(client, key)
         if not field_id:
-            log.warning(f"Could not resolve custom field '{key}' to an ID, skipping")
-            continue
+            log.error(f"Could not resolve custom field '{key}' to an ID, skipping")
+            raise ValueError(
+                f"Could not resolve custom field '{key}' to an ID, skipping"
+            )
```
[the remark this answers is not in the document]
Bala-Sakabattula: Okay, I removed the log.error

tests/test_downstream_issue.py
```diff
@@ -1933,15 +1939,58 @@ def test_create_jira_issue_epic_link_field_not_found(
...
+        ]
+        self.mock_config["sync2jira"]["default_jira_fields"][
+            "storypoints"
+        ] = "Story Points"
+
+        # Set up mock - storypoints field identifier doesn't exist in JIRA
+        mock_client.fields.return_value = [
+            {"name": "Priority", "id": "priority"},
+            # Note: The field from default_jira_fields.storypoints doesn't exist
+        ]
```
webbnh: Again, I'm not a fan of what Black does to line comments, so I would move the note to the block comment:
```suggestion
        # Set up mock - storypoints field identifier doesn't exist in JIRA
        # Note: The field from default_jira_fields.storypoints doesn't exist
        mock_client.fields.return_value = [{"name": "Priority", "id": "priority"}]
```
Again, this is more readable, I think the application of the note is still clear, and it uses about half as many lines.

PR #405, opened 2026-01-09
Issue can be created from PR and handling if  github_project_fields is None 

Bala-Sakabattula:
### Summary

> Added support to automatically create a Jira issue for a Pull Request when enabled via project configuration. This is controlled by a new config flag and improves flexibility for teams that want PR-driven Jira tracking.

### Changes

> 
> Introduced a new configuration option:
> 
> `create_pr_issue: true`
> 
> 
> When set, a Jira issue will be created for the corresponding GitHub PR.
> 
> Added handling for cases where github-project-fields is set to None in the configuration, preventing errors and allowing smoother execution.

Testing

> Tested locally and verified successful Jira issue creation.
> 
> Confirmed issue creation in the dev Jira instance:
> 
> Jira issue: https://redhat-development.atlassian.net/browse/SYNC2JIRA-22
> 
> GitHub PR: https://github.com/Bala-Sakabattula/demo-sync2jira/pull/4

### DIFF
--- a/.tekton/build-pipeline.yaml
+++ b/.tekton/build-pipeline.yaml
@@ -134,7 +134,7 @@ spec:
       - name: name
         value: git-clone-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:0a89e1a6304076525e9766f63a4cd006763d21d5aca6863281fc427537a23c6f
+        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:56f65a16d3d0485c64ad85af2c1f3e9b0bb4d02d63f2fd0ebb9498d219ca723d
       - name: kind
         value: task
       resolver: bundles
@@ -163,7 +163,7 @@ spec:
       - name: name
         value: prefetch-dependencies-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.2@sha256:3e5e834290a1ed57fd14c0082e5a10789c8fe382ed682ef7f981475a7b316b49
+        value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.2@sha256:4c9ff416bfd127e1f960bd0218127c7e198dbd15827c1a8bf58ac5eb023dd9e2
       - name: kind
         value: task
       resolver: bundles
@@ -210,7 +210,7 @@ spec:
       - name: name
         value: buildah-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-buildah-oci-ta:0.7@sha256:2de614f405527e779534a5d1a1293a528c482aa6abebc8ea158ad47e4be5dea4
+        value: quay.io/konflux-ci/tekton-catalog/task-buildah-oci-ta:0.7@sha256:7d5818e082e5534cf63946c1a1d380c0ee6b10b5c915340368c9ca081b97c02a
       - name: kind
         value: task
       resolver: bundles
@@ -241,7 +241,7 @@ spec:
       - name: name
         value: build-image-index
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-build-image-index:0.2@sha256:39561ac43e325159497c10c0284cf61dfddf39e39100ca5e3df6b73c5d96db8b
+        value: quay.io/konflux-ci/tekton-catalog/task-build-image-index:0.2@sha256:985d1efe861b02524a7679ecd855624b3d4e3a2e835b6f8a97ec7d135898ec0b
       - name: kind
         value: task
       resolver: bundles
@@ -267,7 +267,7 @@ spec:
       - name: name
         value: source-build-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-source-build-oci-ta:0.3@sha256:4abb2dbc9dcfad52d56b490a2f25f99989a2cb2bbd9881223025272db60fd75e
+        value: quay.io/konflux-ci/tekton-catalog/task-source-build-oci-ta:0.3@sha256:c35ba219390d77a48ee19347e5ee8d13e5c23e3984299e02291d6da1ed8a986c
       - name: kind
         value: task
       resolver: bundles
@@ -293,7 +293,7 @@ spec:
       - name: name
         value: deprecated-image-check
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-deprecated-image-check:0.5@sha256:462baed733dfc38aca5395499e92f19b6f13a74c2e88fe5d86c3cffa2f899b57
+        value: quay.io/konflux-ci/tekton-catalog/task-deprecated-image-check:0.5@sha256:1cf21de671be4c97d4973b60c09c912997cd15b65c30b93a07eff1b24f43a1f8
       - name: kind
         value: task
       resolver: bundles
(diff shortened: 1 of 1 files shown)


sync2jira/downstream_pr.py
```diff
@@ -234,3 +252,61 @@ def update_jira(client, config, pr):
...
+        pr.upstream,
+        pr.comments,
+        pr.priority,
+        pr.content,
+        pr.reporter,
+        pr.assignee,
+        pr.status,
+        pr.id,
+    )
+    # exit(1)
```
webbnh: I don't think you want to merge these lines into `main`.

sync2jira/downstream_pr.py
```diff
@@ -234,3 +252,61 @@ def update_jira(client, config, pr):
...
+        url=pr.url,
+        upstream=pr.upstream,
+        comments=pr.comments,
+        config=config,
+        tags=[],  # PRs don't have tags in the same way
+        fixVersion=[],
+        priority=pr.priority,
+        content=pr.content or f"PR: {pr.url}",
+        reporter=(
+            {"fullname": pr.reporter} if isinstance(pr.reporter, str) else pr.reporter
```
webbnh: I wonder if this expression should be more like,
```suggestion
            pr.reporter if isinstance(pr.reporter, dict) else {"fullname": str(pr.reporter)}
```
That is, if the provided value is already a dictionary, then we can just use it (and we hope that it contains the proper key and value...); otherwise, we construct a dictionary with the value coerced to be a string -- if it was already a string, then the coercion is trivial but if it was something else then we avoid the pitfall of using it as is (when what Jira wants is a `dict` with a `str`).

sync2jira/upstream_issue.py
```diff
@@ -226,6 +226,11 @@ def add_project_values(issue, upstream, headers, config):
     issue["priority"] = None
     issuenumber = issue["number"]
     github_project_fields = upstream_config["github_project_fields"]
+    if not github_project_fields:
+        log.debug(
+            "github_project_fields is None or empty, skipping project field updates"
+        )
+        return
```
webbnh: Should we have unit test scenarios for each of the early-exit cases?

tests/test_downstream_pr.py
```diff
@@ -370,3 +370,207 @@ def test_update_transition(self, mock_d_issue):
...
+        call_args = mock_issue_class.call_args
+        self.assertEqual(
+            call_args[1]["content"], "PR: https://github.com/test/repo/pull/1"
+        )
+
+    @mock.patch(PATH + "d_issue._create_jira_issue")
+    @mock.patch("sync2jira.intermediary.Issue")
+    def test_create_jira_issue_from_pr_reporter_dict(
+        self, mock_issue_class, mock_create_jira_issue
+    ):
```
webbnh: There's a lot of duplication between `test_create_jira_issue_from_pr()`, `test_create_jira_issue_from_pr_no_content()`, and `test_create_jira_issue_from_pr_reporter_dict()`.  You might want to factor the common setup into a helper function to address that.  You might want to do something similar for the assertions which are (or would be) common across all three (i.e., to prove that the different scenarios affect _only_ the targeted subset of the fields).

tests/test_downstream_pr.py
```diff
@@ -370,3 +370,207 @@ def test_update_transition(self, mock_d_issue):
...
+        self.assertEqual(call_args[1]["source"], "github")
+        self.assertEqual(call_args[1]["title"], "Test PR Title")
+        self.assertEqual(call_args[1]["url"], "https://github.com/test/repo/pull/1")
+        self.assertEqual(call_args[1]["upstream"], "test/repo")
+        self.assertEqual(call_args[1]["status"], None)
+        self.assertEqual(call_args[1]["reporter"], {"fullname": "testuser"})
+
+        # Assert _create_jira_issue was called
+        mock_create_jira_issue.assert_called_once()
+        self.assertEqual(result, mock_created_issue)
```
webbnh: Do we really care that `_create_jira_issue()` was called (at all), given that we check the value that the CUT returned?

sync2jira/downstream_pr.py
```diff
@@ -177,9 +177,18 @@ def sync_with_jira(pr, config):
...
-        log.info(f"[PR] No match found for {pr.title}")
-        return None
+        if create_pr_issue:
+            log.info(
+                f"No JIRA key found in PR {pr.title}, but 'create_pr_issue' is enabled. "
+                f"Will create new JIRA issue."
+            )
+        else:
+            log.info(f"[PR] No match found for {pr.title}")
+            return None
```
webbnh: While Python doesn't really care, this function doesn't return a value (e.g., in the success case), so you should omit the `None` from this `return` statement.  (Ditto at line 178, if you want to fix that, too.)

sync2jira/downstream_pr.py
```diff
@@ -264,27 +255,18 @@ def _create_jira_issue_from_pr(client, pr, config):
     :returns: Created JIRA issue
     :rtype: jira.resources.Issue
     """
-    from sync2jira.intermediary import Issue
+
+    pr_content = pr.content or f"PR: {pr.url}"
+    if pr.downstream.get("issue_updates"):
+        if (
+            pr.source == "github"
```
webbnh: temp

sync2jira/downstream_pr.py
```diff
@@ -219,7 +228,16 @@ def update_jira(client, config, pr):
         pr.jira_key = matcher(pr.content, pr.comments)
 
     if not pr.jira_key:
-        log.info("No JIRA key found in PR, skipping.")
+        create_pr_issue = pr.downstream.get("create_pr_issue", False)
+        if create_pr_issue:
+            log.info(
+                f"No JIRA key found in PR {pr.url}, but 'create_pr_issue' is enabled. "
+                f"Creating new JIRA issue for PR."
+            )
```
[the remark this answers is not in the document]
webbnh: The concern is, now that we have `sync2jira` on a fairly robust footing, the log easily grows to 100k lines between redeployments...so, I'm resistant to anything which makes it larger unless it's providing good value for the cost.

tests/test_downstream_pr.py
```diff
@@ -370,3 +369,212 @@ def test_update_transition(self, mock_d_issue):
...
+        self.mock_pr.priority = None
+        self.mock_pr.content = "PR description"
+        self.mock_pr.reporter = "testuser"
+        self.mock_pr.assignee = []
+        self.mock_pr.status = None
+        self.mock_pr.id = "1"
+        self.mock_pr.downstream = {"project": "TEST", "type": "Task"}
+        # Apply any overrides
+        for key, value in overrides.items():
+            setattr(self.mock_pr, key, value)
```
webbnh: Nice!  🥇

webbnh (commented):
This is a very nice update.  Unfortunately, it needs one more pass:

- [ ] The pandoc support isn't quite hooked up.
- [ ] Now that `sync_with_jira()` consistently doesn't return a value, the unit test for it needs to be changed to match
- [ ] I think we want a negative assertion in the second `update_jira()` scenario.

and there are a couple of smaller items for you to consider.

Also, there are a couple of unaddressed comments ([1](https://github.com/release-engineering/Sync2Jira/pull/405#discussion_r2683208326), [2](https://github.com/release-engineering/Sync2Jira/pull/405#discussion_r2683211908)) from my first review.

sync2jira/downstream_pr.py
Bala-Sakabattula: Okay! from next time I will take care of this 🙂

tests/test_upstream_issue.py
```diff
@@ -725,3 +725,73 @@ def test_get_current_project_node(self):
...
+        result = u.add_project_values(
+            issue=mock_issue,
+            upstream="org/repo",
+            headers={},
+            config=self.mock_config,
+        )
+
+        # Assert requests.post was not called (early exit occurred)
+        mock_requests_post.assert_not_called()
+        self.assertIsNone(result)
```
webbnh: Same comment as previous.
```suggestion
        scenarios = (
            # Test case 1: github_project_fields is None
            (None, ["github_project_fields"]),
            # Test case 2: github_project_fields is empty dict
            ({}, ["github_project_fields"]),
            # Test case 3: "github_project_fields" not in issue_updates
            ({"storypoints": {"gh_field": "Estimate"}}, []),
        )

        for gpf, iu in scenarios:
                upstream_config["github_project_fields"] = gpf
                upstream_config["issue_updates"] = ["comments", "title"] + iu

                result = u.add_project_values(
                    issue=mock_issue,
                    upstream="org/repo",
                    headers={},
                    config=self.mock_config,
                )

                # Assert requests.post was not called (early exit occurred)
                mock_requests_post.assert_not_called()
                self.assertIsNone(result)

                # Reset mock
                mock_requests_post.reset_mock()
```

PR #433, opened 2026-02-18
client session updation

Bala-Sakabattula:
I updated the Jira client call used for checking service availability from `session()` to `server_info()` as [the `session()` function is depreciated and will be removed soon](https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-basic-auth-and-cookie-based-auth/).


webbnh (approved):
Looks good to me.

PR #448, opened 2026-04-02
JIRA cloud issue creation and availability issue

Bala-Sakabattula:
Created this PR for solving the following issue:

1. After jira cloud issue creation instantly if you search for through JQL it won't return so for that we have used GET call as an alternative if the key is present in cache or in snowflake but not in JQL ...

2. As previously we are logging issue Id but it isn't that useful so we are now chnaged to the log the JQL contents it will help to debug in better way.

3. Updated to use displayName instead of name in `_update_Assignee` and `check_comments_for_duplicate`

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -23,7 +23,7 @@ import logging
 import operator
 import os
 import re
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Tuple, Union
 import unicodedata
 
 from dotenv import load_dotenv
@@ -316,22 +316,34 @@ def get_existing_jira_issue(client, issue, config):
     :rtype: JIssue or None
     """
 
-    # Fetch the Jira issue objects using the key list.
-    jql = _get_existing_jira_issue_query(issue)
+    issue_keys, jql = _get_existing_jira_issue_query(issue)
     if not jql:
         return None
     results: ResultList[JIssue] = client.search_issues(jql)
     if not results:
-        # It is _possible_ that the downstream issue exists (i.e., we did hit
-        # it in the cache or in Snowflake) but the Jira search cannot find it.
-        # (This happens when the issue has been archived.)  Fail as gracefully
-        # as we can here, but our caller will probably create it again.
-        log.warning(
-            "Previously-existing downstream issue %s not found for upstream issue %s.",
-            issue.id,
-            issue.url,
-        )
-        return None
+        # JQL/search index can lag right after create, or hide archived issues.
+        # Resolve by issue key (direct GET) before giving up and duplicating.
+        results = ResultList[JIssue]()
+        for key in issue_keys:
+            try:
+                found = client.issue(key)
+            except JIRAError:
+                continue
+            log.info(
+                "JQL missed key %s for upstream %s; using direct issue fetch.",
+                key,
+                issue.url,
+            )
+            results = ResultList[JIssue]((found,))
+            break
+        if not results:
+            log.warning(
+                "Downstream issue not found for upstream %s after JQL %r and direct fetch for keys %s.",
+                issue.url,
+                jql,
+                issue_keys,
+            )
+            return None
 
     # If there is more than one issue, remove duplicates and filter the list
     # down to one.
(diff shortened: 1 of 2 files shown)


sync2jira/downstream_issue.py
```diff
@@ -316,22 +316,34 @@ def get_existing_jira_issue(client, issue, config):
     :rtype: JIssue or None
     """
 
-    # Fetch the Jira issue objects using the key list.
-    jql = _get_existing_jira_issue_query(issue)
+    issue_keys, jql = _get_existing_jira_issue_query(issue)
     if not jql:
         return None
     results: ResultList[JIssue] = client.search_issues(jql)
```
webbnh, line 322: I don't think it's worthwhile to return a tuple.  Instead, I suggest refocusing (and renaming) the subroutine and moving the query creation to the caller:
```suggestion
    issue_keys = _get_existing_jira_issue_keys(issue)
    if not issue_keys:
        return None
    jql = f"key in ({','.join(issue_keys)})
    results: ResultList[JIssue] = client.search_issues(jql")
```

sync2jira/downstream_issue.py
```diff
@@ -316,22 +316,34 @@ def get_existing_jira_issue(client, issue, config):
...
-        # as we can here, but our caller will probably create it again.
-        log.warning(
-            "Previously-existing downstream issue %s not found for upstream issue %s.",
-            issue.id,
-            issue.url,
-        )
-        return None
+        # JQL/search index can lag right after create, or hide archived issues.
+        # Resolve by issue key (direct GET) before giving up and duplicating.
+        results = ResultList[JIssue]()
```
webbnh, line 326: You don't need to assign `results` here -- the `if` statement at line 323 ensures that it's an empty list already.

sync2jira/downstream_issue.py
```diff
@@ -316,22 +316,34 @@ def get_existing_jira_issue(client, issue, config):
...
+            except JIRAError:
+                continue
+            log.info(
+                "JQL missed key %s for upstream %s; using direct issue fetch.",
+                key,
+                issue.url,
+            )
+            results = ResultList[JIssue]((found,))
+            break
+        if not results:
```
webbnh, line 339: Rather than using `if not results`, I think it would be more graceful to just use `else` -- if we don't `break` out of the `for` loop, then we'll execute the `else` block.

sync2jira/downstream_issue.py
```diff
@@ -445,7 +459,11 @@ def check_comments_for_duplicate(client, result, username):
     """
     for comment in client.comments(result):
         search = re.search(r"Marking as duplicate of (\w*)-(\d*)", comment.body)
-        if search and comment.author.name == username:
+        author = comment.author
+        author_label = getattr(author, "displayName", None) or getattr(
+            author, "name", None
+        )
```
webbnh, line 465: Alternatively,
```suggestion
        author_label = getattr(author, "displayName", getattr(author, "name", None))
```

sync2jira/downstream_issue.py
```diff
@@ -1139,7 +1157,9 @@ def _update_assignee(client, existing, issue, overwrite):
         else:
             # Without an upstream owner, update only if the downstream is not
             # assigned to the project owner.
-            update = issue.downstream.get("owner") != existing.fields.assignee.name
+            update = (
+                issue.downstream.get("owner") != existing.fields.assignee.displayName
+            )
```
webbnh, line 1162: Why do we use `getattr()` to access `displayName` at line 463 but use a direct access here?  (Is one of the two "wrong"?)

tests/test_downstream_issue.py
```diff
@@ -287,55 +287,78 @@ def test_get_existing_newstyle(self, mock_client, mock_get_query, mock_filter):
         mock_issue_3.fields = MagicMock()
         mock_issue_3.fields.updated = "2025-11-30T00:00:00.0+0000"
 
+        def _q(keys):
```
webbnh, line 290: Access to this function is scoped to this function, so you don't need the underscore prefix on the name.

tests/test_downstream_issue.py
```diff
@@ -287,55 +287,78 @@ def test_get_existing_newstyle(self, mock_client, mock_get_query, mock_filter):
...
             d.jira_cache = d.UrlCache()  # Clear the cache
-            mock_get_query.return_value = x["jira_results"]
+            mock_get_query.return_value = x["query_return"]
             mock_client.search_issues.return_value = x["search_issues"]
             mock_filter.return_value = x["filter_results"]
+            if x["issue_side_effect"] is not None:
+                mock_client.issue.side_effect = x["issue_side_effect"]
+            else:
+                mock_client.issue.side_effect = None
+                mock_client.issue.return_value = mock_issue_1
```
webbnh, line 361: I believe that you can simplify this:
```suggestion
            mock_client.issue.side_effect = x["issue_side_effect"]
            if not x["issue_side_effect"]:
                mock_client.issue.return_value = mock_issue_1
```
or possibly even
```suggestion
            if not mock_client.issue.side_effect := x["issue_side_effect"]:
                mock_client.issue.return_value = mock_issue_1
```

webbnh (approved):
I'm OK with this as is, since we very much want to stop creating duplicates, but I have a couple of concerns (and a bunch of suggestions).

First, I recommend changing the focus of `_get_existing_jira_issue_query()` (and its name) so that it returns the key list and _not_ the JQL -- the caller can produce the JQL easily enough, and returning a naked tuple is something best to avoid unless there's a compelling need.

Second, you modified two access to use `displayName`, but their respective approaches are inconsistent...which is fine if it's justified, but my instinct is that one of them is "right" and one is "wrong" -- and I don't know which is which....

So, perhaps we could merge this and then you could polish it in a follow-up PR.

PR #450, opened 2026-04-05
Code review fixes for PR #448

Bala-Sakabattula:
Fixes for comments raised in PR #448.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -316,7 +316,7 @@ def get_existing_jira_issue(client, issue, config):
     :rtype: JIssue or None
     """
 
-    issue_keys = _get_existing_jira_issue_query(issue)
+    issue_keys = _get_existing_jira_issue_keys(issue)
     if not issue_keys:
         return None
     jql = f"key in ({','.join(issue_keys)})"
@@ -370,22 +370,23 @@ def get_existing_jira_issue(client, issue, config):
     return results[0]
 
 
-def _get_existing_jira_issue_query(issue: Issue) -> tuple[str, ...]:
+def _get_existing_jira_issue_keys(issue: Issue) -> tuple[str, ...]:
     """
-    Generate a JQL query to find downstream issues corresponding to a given
-    upstream issue.  Return empty tuple if no matches were found in either our local
-    cache or in the Dataverse.
+    Retrieve downstream Jira issue keys corresponding to a given upstream issue.
+
+    The function first checks the local cache; if no cached result is found,
+    it queries Snowflake. Returns empty tuple if no matches are found.
 
     :param sync2jira.intermediary.Issue issue: Issue object
-    :returns: A string containing the JQL query or None if no matches
-    :rtype: Tuple[tuple[str, ...]
+    :returns: A tuple of Jira issue keys, or None if no matches are found
+    :rtype: Optional[Tuple[str, ...]]
     """
     if result := jira_cache.get(issue.url):
         issue_keys = (result,)
     else:
         results = execute_snowflake_query(issue)
         if not results:
-            return None
+            return ()
         issue_keys = tuple(row[0] for row in results)
 
     return issue_keys
@@ -448,7 +449,7 @@ def _jira_user_display_label(user) -> Optional[str]:
     """Best-effort display string for a Jira User (Cloud: displayName, else name)."""
     if not user:
         return None
-    return getattr(user, "displayName", getattr(user, "name", None))
+    return getattr(user, "displayName", None) or getattr(user, "name", None)
 
 
 def check_comments_for_duplicate(client, result, username):
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -444,6 +444,13 @@ def find_username(_issue, config):
     return config["sync2jira"]["jira_username"]
 
 
+def _jira_user_display_label(user) -> Optional[str]:
+    """Best-effort display string for a Jira User (Cloud: displayName, else name)."""
+    if not user:
+        return None
+    return getattr(user, "displayName", getattr(user, "name", None))
+
+
 def check_comments_for_duplicate(client, result, username):
     """
     Checks comment of JIRA issue to see if it has been
@@ -457,8 +464,7 @@ def check_comments_for_duplicate(client, result, username):
     """
     for comment in client.comments(result):
         search = re.search(r"Marking as duplicate of (\w*)-(\d*)", comment.body)
-        author = comment.author
-        author_label = getattr(author, "displayName", getattr(author, "name", None))
+        author_label = _jira_user_display_label(comment.author)
         if search and author_label == username:
             issue_id = search.groups()[0] + "-" + search.groups()[1]
             return client.issue(issue_id)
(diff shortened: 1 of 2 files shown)


sync2jira/downstream_issue.py
```diff
@@ -316,14 +316,14 @@ def get_existing_jira_issue(client, issue, config):
     :rtype: JIssue or None
     """
 
-    issue_keys, jql = _get_existing_jira_issue_query(issue)
-    if not jql:
+    issue_keys = _get_existing_jira_issue_query(issue)
+    if not issue_keys:
         return None
+    jql = f"key in ({','.join(issue_keys)})"
```
webbnh, line 322: `_get_existing_jira_issue_query()` no longer returns a JQL query; you don't want to rename the function?

sync2jira/downstream_issue.py
```diff
@@ -370,27 +370,25 @@ def get_existing_jira_issue(client, issue, config):
...
 
 
-def _get_existing_jira_issue_query(
-    issue: Issue,
-) -> Tuple[tuple[str, ...], Optional[str]]:
+def _get_existing_jira_issue_query(issue: Issue) -> tuple[str, ...]:
     """
     Generate a JQL query to find downstream issues corresponding to a given
-    upstream issue.  Return empty tuple and None if no matches were found in either our local
+    upstream issue.  Return empty tuple if no matches were found in either our local
```
webbnh: The function now returns just the list of keys -- it does _not_ "generate a JQL query" -- so the docstring (new) line 375 should presumably be updated.

sync2jira/downstream_issue.py
```diff
@@ -370,27 +370,25 @@ def get_existing_jira_issue(client, issue, config):
...
-    :rtype: Tuple[tuple[str, ...], Optional[str]]
+    :rtype: Tuple[tuple[str, ...]
     """
     if result := jira_cache.get(issue.url):
         issue_keys = (result,)
     else:
         results = execute_snowflake_query(issue)
         if not results:
-            return (), None
+            return None
```
webbnh: This is contrary to both the function signature type hint and the docstring.  Presumably it should be returning an empty tuple instead of `None`:
```suggestion
        if not results:
            return ()
```

Alternatively, you can presumably just omit the whole `if` statement:  if `execute_snowflake_query()` returns an empty list, then the construction at (new) line 389 will produce an empty tuple, and we'll return the same result.

sync2jira/downstream_issue.py
```diff
@@ -446,6 +444,13 @@ def find_username(_issue, config):
     return config["sync2jira"]["jira_username"]
 
 
+def _jira_user_display_label(user) -> Optional[str]:
+    """Best-effort display string for a Jira User (Cloud: displayName, else name)."""
+    if not user:
+        return None
+    return getattr(user, "displayName", getattr(user, "name", None))
```
webbnh: @Bala, this encapsulation is an excellent step!

Nevertheless, what do you think of Qodo's [second comment](https://github.com/release-engineering/Sync2Jira/pull/450#issuecomment-4188900541) from yesterday?  _Can_ the display name be an empty string (or an explicit `None`), and, if so, do we want to  try to fall back to using `"name"` when it is?

(While I like and prefer the nested `getattr()` call, if we need/want to cover that case, then your original code was better, modulo what Black did to it....)

tests/test_downstream_issue.py
```diff
@@ -1715,6 +1706,21 @@ def test_find_username(self):
...
+        self.assertIsNone(d._jira_user_display_label(None))
+        self.assertIsNone(d._jira_user_display_label(""))
+        self.assertEqual(
+            d._jira_user_display_label(types.SimpleNamespace(displayName="Alice")),
+            "Alice",
+        )
+        self.assertEqual(
+            d._jira_user_display_label(types.SimpleNamespace(name="bob_only")),
+            "bob_only",
+        )
```
webbnh, line 1722: You're missing the case where the parameter is not false-y but contains _neither_ `displayName` nor `name`.  Also, it would be good to have a case where the parameter contains _both_ fields, and we show that it returns the correct one.  (That is, since there are two possible field-inputs, there should be four cases:  "A", "B", both, and neither.  🙂)

webbnh (commented):
Looks good, but it needs a little polish.

The docstring for `_get_existing_jira_issue_query()` needs to be updated, and either it or the actual `return` need to be brought into alignment.  (Qodo commented on this, too.)  And, I recommend changing the function's name.

The encapsulation of `_jira_user_display_label()` test is excellent, but the test for it is missing a case or two.  Also, Qodo has a comment about its code that we should consider.

sync2jira/downstream_issue.py
Bala-Sakabattula: I am not sure about it so should  we go back to old type ?? 
```
getattr(user, "displayName", None) or getattr( user, "name", None)
```

webbnh: Yeah, I suppose so.  But, thanks to this encapsulation, you only have to change it in one place!  🙂

The trick is structuring it in a way which doesn't tempt Black to spread it over extra lines making it _less_ readable.  😛

sync2jira/downstream_issue.py
```diff
@@ -370,22 +370,23 @@ def get_existing_jira_issue(client, issue, config):
...
-    cache or in the Dataverse.
+    Retrieve downstream Jira issue keys corresponding to a given upstream issue.
+
+    The function first checks the local cache; if no cached result is found,
+    it queries Snowflake. Returns empty tuple if no matches are found.
 
     :param sync2jira.intermediary.Issue issue: Issue object
-    :returns: A string containing the JQL query or None if no matches
-    :rtype: Tuple[tuple[str, ...]
+    :returns: A tuple of Jira issue keys, or None if no matches are found
```
webbnh: Replace "or None" with "empty".  😇

sync2jira/downstream_issue.py
```diff
@@ -370,22 +370,23 @@ def get_existing_jira_issue(client, issue, config):
...
+    :rtype: Optional[Tuple[str, ...]]
```
webbnh: The return is _not_ `Optional` (it will be a tuple, empty not `None`, if there are no matches found).

sync2jira/downstream_issue.py
webbnh: You responded with "👍" but you didn't make the change...I hope that that was what you meant.

webbnh (dismissed):
Good to go, but there are three nits...let me know if you don't want to address them and I'll resolve the comments.

sync2jira/downstream_issue.py
Bala-Sakabattula: > This is contrary to both the function signature type hint and the docstring. Presumably it should be returning an empty tuple instead of None

I made this change I returned empty tuple and changed the docstring as well..

webbnh: Quite right -- my apologies.

In my haste, I mistook what "the change" was and wasn't clear about that in my follow-up.  What I meant was to point out that you opted to retain the `if` statement (and to make it explicitly return an empty tuple) rather than removing it and letting execution continue on and _construct_ an empty tuple (from the empty list which was returned); I wanted to confirm that that was what you intended.

At this point, lets just go with what you have.  🙂

webbnh (approved):
👍

PR #24, opened 2019-12-11
Fix issue when GitHub pull-request closes an issue

sidpremkumar:
We have an issue where GitHub pull requests can close upstream issues (i.e. "Fixes: #-1" when merged would close issue #-1). This PR fixes that bug.

PR #452, opened 2026-04-10
Support Single Select field mapping for storypoints

istein1:
## Problem

When using GitHub Project **Single Select** fields with numeric option mappings for storypoints, sync2jira fails with:
```
Error while processing storypoints for issue: 'number'
```

This occurs because the code only supports **Number-type** fields and tries to read `item["number"]`, which doesn't exist for Single Select fields.

## Configuration Example

```python
github_project_fields={
    'storypoints': {
        'gh_field': 'Size',
        'options': {
            '🐇 Small': 5,
            '🐂 Medium': 8,
            '🦑 Large': 23,
        }
    }
}
```

## Solution

This PR adds support for Single Select fields with numeric mappings by:
1. Checking if the storypoints config has an `options` mapping
2. For Single Select fields: reading `item["name"]` and mapping it via the options dict
3. For Number fields: reading `item["number"]` directly (preserves existing behavior)
4. Ensuring the final value is always cast to `int` for Jira compatibility

## Testing

- Tested with migtools/crane repository using GitHub org-level project #24
- Single Select "Size" field with emoji options now correctly maps to numeric storypoints in Jira
- Existing Number-field configurations continue to work (backward compatible)

## Related

This enables configurations like those already used in other projects (packit, cpt, tmt, etc.) but with more flexible field types.

### DIFF
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -317,7 +317,8 @@ def add_project_values(issue, upstream, headers, config):
                         issue["storypoints"] = int(sp_number)
                     except (ValueError, TypeError) as err:
                         log.info(
-                            "Error while processing storypoints for issue %s/%s#%s: %s",
+                            "Error converting Single Select storypoints value '%s' to int for issue %s/%s#%s: %s",
+                            sp_number,
                             orgname,
                             reponame,
                             issuenumber,
@@ -329,7 +330,8 @@ def add_project_values(issue, upstream, headers, config):
                     issue["storypoints"] = int(item["number"])
                 except (ValueError, TypeError, KeyError) as err:
                     log.info(
-                        "Error while processing storypoints for issue %s/%s#%s: %s",
+                        "Error converting Number field storypoints value '%s' to int for issue %s/%s#%s: %s",
+                        item.get("number", "missing"),
                         orgname,
                         reponame,
                         issuenumber,
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -292,7 +292,16 @@ def add_project_values(issue, upstream, headers, config):
         sp_field = github_project_fields.get("storypoints", {}).get("gh_field")
         if gh_field_name == sp_field:
             try:
-                issue["storypoints"] = int(item["number"])
+                # Check if there's an options mapping (for Single Select fields)
+                sp_options = github_project_fields.get("storypoints", {}).get("options")
+                if sp_options:
+                    # Single Select field - get name and map it
+                    sp_value = item.get("name")
+                    if sp_value and sp_value in sp_options:
+                        issue["storypoints"] = int(sp_options[sp_value])
+                else:
+                    # Number field - get number directly
+                    issue["storypoints"] = int(item["number"])
             except (ValueError, KeyError) as err:
                 log.info(
                     "Error while processing storypoints for issue %s/%s#%s: %s",
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -291,34 +291,50 @@ def add_project_values(issue, upstream, headers, config):
             continue
         sp_field = github_project_fields.get("storypoints", {}).get("gh_field")
         if gh_field_name == sp_field:
-            try:
-                # Check if there's an options mapping (for Single Select fields)
-                sp_options = github_project_fields.get("storypoints", {}).get("options")
-                if sp_options:
-                    # Single Select field - get name and map it
-                    sp_value = item.get("name")
-                    if sp_value and sp_value in sp_options:
-                        mapped_value = sp_options[sp_value]
-                        issue["storypoints"] = int(mapped_value)
-                    else:
+            # Check if there's an options mapping (for Single Select fields); if
+            # so, convert...
+            sp_options = github_project_fields.get("storypoints", {}).get("options")
+            if sp_options:
+                # Single Select field - get name and map it
+                sp_value = item.get("name")
+                if not sp_value:
+                    log.warning(
+                        "No Single Select name found for storypoints options in message for issue %s/%s#%s",
+                        orgname,
+                        reponame,
+                        issuenumber,
+                    )
+                elif (sp_number := sp_options.get(sp_value)) is None:
+                    log.info(
+                        "Storypoints value '%s' not found in options mapping for issue %s/%s#%s",
+                        sp_value,
+                        orgname,
+                        reponame,
+                        issuenumber,
+                    )
+                else:
+                    try:
+                        issue["storypoints"] = int(sp_number)
+                    except (ValueError, TypeError) as err:
                         log.info(
-                            "Storypoints value '%s' not found in options mapping for issue %s/%s#%s",
-                            sp_value,
+                            "Error while processing storypoints for issue %s/%s#%s: %s",
                             orgname,
                             reponame,
                             issuenumber,
+                            err,
                         )
-                else:
-                    # Number field - get number directly
+            else:
+                # Number field - get number directly
+                try:
                     issue["storypoints"] = int(item["number"])
-            except (ValueError, TypeError, KeyError) as err:
-                log.info(
-                    "Error while processing storypoints for issue %s/%s#%s: %s",
-                    orgname,
-                    reponame,
-                    issuenumber,
-                    err,
-                )
+                except (ValueError, TypeError, KeyError) as err:
+                    log.info(
+                        "Error while processing storypoints for issue %s/%s#%s: %s",
+                        orgname,
+                        reponame,
+                        issuenumber,
+                        err,
+                    )
             continue
 
 
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -289,12 +289,12 @@ def add_project_values(issue, upstream, headers, config):
         if gh_field_name == prio_field:
             issue["priority"] = item.get("name")
             continue
-        sp_field = github_project_fields.get("storypoints", {}).get("gh_field")
+        sp_dict = github_project_fields.get("storypoints", {})
+        sp_field = sp_dict.get("gh_field")
         if gh_field_name == sp_field:
             # Check if there's an options mapping (for Single Select fields); if
             # so, convert...
-            sp_options = github_project_fields.get("storypoints", {}).get("options")
-            if sp_options:
+            if sp_options := sp_dict.get("options"):
                 # Single Select field - get name and map it
                 sp_value = item.get("name")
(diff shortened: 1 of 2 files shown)


sync2jira/upstream_issue.py
```diff
@@ -292,7 +292,16 @@ def add_project_values(issue, upstream, headers, config):
...
+                sp_options = github_project_fields.get("storypoints", {}).get("options")
+                if sp_options:
+                    # Single Select field - get name and map it
+                    sp_value = item.get("name")
+                    if sp_value and sp_value in sp_options:
+                        issue["storypoints"] = int(sp_options[sp_value])
+                else:
+                    # Number field - get number directly
+                    issue["storypoints"] = int(item["number"])
             except (ValueError, KeyError) as err:
```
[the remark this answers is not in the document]
istein1: Issue fixed.

sync2jira/upstream_issue.py
```diff
@@ -292,8 +292,26 @@ def add_project_values(issue, upstream, headers, config):
...
+                            "Storypoints value '%s' not found in options mapping for issue %s/%s#%s",
+                            sp_value,
+                            orgname,
+                            reponame,
+                            issuenumber,
+                        )
+                else:
+                    # Number field - get number directly
+                    issue["storypoints"] = int(item["number"])
+            except (ValueError, TypeError, KeyError) as err:
```
webbnh: There's an awful lot of logic inside the `try` block which should never fail.  I recommend moving it to before the `try`.

Also, your code combines two failure cases into one, which I recommend separating.

Finally, this might be a good spot for the "walrus" operator, so I suggest something like this:
```suggestion
            # Check if there's an options mapping (for Single Select fields):  if not, use the
            # number; if so, get the option name and fetch the mapping.  Either way, convert
            # the resulting value to an integer for Jira.
            if not (sp_options := github_project_fields.get("storypoints", {}).get("options")):
                sp_number = item["number"]
            elif not (sp_value := item.get("name")):
                log.warning(
                    "No Single Select name found for `storypoint.options` in message for issue %s/%s#%s",
                    orgname,
                    reponame,
                    issuenumber,
                )
                continue
            elif not (sp_number := sp_options.get(sp_value)):
                log.info(
                    "Storypoints value '%s' not found in options mapping for issue %s/%s#%s",
                    sp_value,
                    orgname,
                    reponame,
                    issuenumber,
                )
                continue
            try:
                issue["storypoints"] = int(sp_number)
            except (ValueError, TypeError, KeyError) as err:
```
The missing SS name should (I believe) never happen, so it's reported as a warning-level message (but, I think it's still worth checking).

webbnh (changes requested):
Thanks so much for the contribution!

And, thanks for paying attention to Qodo's comments.  I've enabled testing for this PR, and it confirms the last of them:  the CI is failing due to lack of test coverage for the new code.

I've posed a suggestion for your change which reduces the burden on the `try` block and hopefully better addresses Qodo's second comment.

sync2jira/upstream_issue.py
webbnh: Hmmm, my suggestion might need a tweak:  if the mapped value is a numeric zero, it will complain -- whether that is bad or not probably depends on whether you think zero is a valid story point value.  🙂

Perhaps that line should be something more like,
```python
            elif (sp_number := sp_options.get(sp_value)) is None:
```
or
```python
            elif not (sp_number := sp_options.get(sp_value)) and sp_number != 0:
```

istein1: Thank you @webbnh for the review!
Added a fix.

sync2jira/upstream_issue.py
```diff
@@ -291,16 +291,50 @@ def add_project_values(issue, upstream, headers, config):
...
+                try:
+                    issue["storypoints"] = int(item["number"])
+                except (ValueError, TypeError, KeyError) as err:
+                    log.info(
+                        "Error while processing storypoints for issue %s/%s#%s: %s",
+                        orgname,
+                        reponame,
+                        issuenumber,
+                        err,
+                    )
```
webbnh, line 339: These two logging calls use the same text, which means that, during a post-mortem analysis, it might be hard to figure out what happened.  If you could make them distinct from each other, that would be a help.
 
Also, I'm not sure whether the `err` value will include the value of the argument to `int()`; if it _doesn't_, it would be very handy to have that in the logged message, as well.  (It should be easy to see when you add testing for this code.)

webbnh (changes requested):
The code change looks very good (I have just one small request).

However, the CI is failing due to a lack of test coverage.  If you could [add four or so cases](https://github.com/istein1/Sync2Jira/blob/85975df27a2d7d87dcdc912493d0462538927191/tests/test_upstream_issue.py#L748-L755) to `test_add_project_values_early_exit()`, that should satisfy it.

webbnh:
> If you could [add four or so cases](https://github.com/istein1/Sync2Jira/blob/85975df27a2d7d87dcdc912493d0462538927191/tests/test_upstream_issue.py#L748-L755) to `test_add_project_values_early_exit()`, that should satisfy it.

Actually, that's not a good fit.  Instead, we need a new test (similar to, but not quite the same as, `test_add_project_values_early_exit()`.

sync2jira/upstream_issue.py
```diff
@@ -291,16 +291,50 @@ def add_project_values(issue, upstream, headers, config):
...
-                    "Error while processing storypoints for issue %s/%s#%s: %s",
-                    orgname,
-                    reponame,
-                    issuenumber,
-                    err,
-                )
+            # Check if there's an options mapping (for Single Select fields); if
+            # so, convert...
+            sp_options = github_project_fields.get("storypoints", {}).get("options")
+            if sp_options:
```
webbnh: Line 293 assures us that there _is_ a `"storypoints"` key in `github_pojrect_fields`, so, at line 296, we can use `[]` access instead of `.get()` to fetch it.  But, between lines 292 and 296, we're fetching the `"storypoints"` value twice, so we might as well cache it in a local variable instead.  Finally, fetching `sp_options` looks like a good place for the walrus operator.

```python
        sp_dict = github_project_fields.get("storypoints", {})
        sp_field = sp_dict.get("gh_field")
        if gh_field_name == sp_field:
            # Check if there's an options mapping (for Single Select fields); if
            # so, convert...
            if sp_options := sp_dict.get("options"):
```

sync2jira/upstream_issue.py
istein1: Thanks @webbnh ,
This is now fixed and the the calls are distincted.

istein1:
@webbnh , Tests are now added.

webbnh (commented):
Thanks for the tests, @istein1!

> All tests pass with 82.58% coverage (exceeds 70% requirement).

Tell Claude that the actual requirement is that coverage _not decrease_, and so the current threshold is about 83.8%.  😏 

> - Add comprehensive test coverage for all error handling scenarios
>
> Test coverage includes:
> - Single Select field with invalid mapped value
> - Number field with invalid number value
> - Single Select field with missing name
> - Single Select field with unmapped value

I don't know that I would call that "comprehensive" (but, I know how enthusiastic Claude can be 😆).  Here's the list that I came up with:

- Test case 1: No `"storypoints"` field in `github_project_fields`
- Test case 2a: No `"gh_field"` field in `github_project_fields["storypoints"]`
- Test case 2b: `item["fieldName"]["name"]` != `github_project_fields["storypoints"]["gh_field"]`
- (Test cases 3: No `"options"` in `github_project_fields["storypoints"]`)
- Test case 3a: `item["number"]` yields `ValueError`
- Test case 3b: `item["number"]` yields `TypeError`
- Test case 3c: `item["number"]` yields `KeyError`?
- Test case 3d: `item["number"]` yields something else?
- Test case 4: `github_project_fields["storypoints"]["options"]` is empty
- Test case 5: no `"name"` in `item`
- Test case 6: `sp_value` not in `sp_options`
- Test case 7a: `sp_value` yields `ValueError`
- Test case 7b: `sp_value` yields `TypeError`
- Test case 7c: `sp_value` yields something else?

So, it looks like Claude hit 7a, 3a, 5, and 6.  We can probably get by without the other variants of cases 3 and 7, but exercising 1, 2a, and 4 would probably be enough.

And, while I was looking at test possibilities, I noted another code tweak and commented below.

One final thought, since you've fired up Claude:  the test code is kind of cumbersome, in that it does the same set-up boilerplate over and over, which makes it harder to pick out what is different between the scenarios.  Could you ask Claude to make the test table-driven (like `test_add_project_values_early_exit()`) and see if it can come up with something more focused?

Alternatively, the list of scenarios that I've presented above is cumulative -- each scenario differs from the previous by adding one more piece of state -- so the test could be constructed similarly, with a single piece being added and then asserted, rather than reconstructing all the state from scratch each time.

Thanks!

istein1:
Thanks @webbnh for the extensive review !!
Added tests and a fix for the last comment.

webbnh (dismissed):
Oh, so close!  The code looks fine to me, now, but Black objects to the formatting.  (And, annoyingly, that prevented us from seeing whether the coverage passed the mark.)  Let me know if you don't know how to fix this.

I believe the two Konflux policy checks are our problem, not yours.  I just merged a Konflux update PR, which I'm hoping will address those, but, you'll need to rebase this branch, and then hopefully the tests will pass.

webbnh:
It's really frustrating not being able to see the annotated source code coverage, but Coveralls has some sort of[ chronic infrastructure problem](https://github.com/lemurheavy/coveralls-public/issues/1757) that they haven't been able to trace.

In any case, Coveralls is convinced that the coverage of the new code added in this PR is still wanting:

> 5 of 16 new or added lines in 1 file covered. (31.25%)

I just pulled your branch and ran the coverage analysis locally and it really doesn't look good:
Screenshot 2026-04-14 at 5 10 54 PM

So, now, we need to figure out what Claude's tests are _actually_ doing....  😞 

A quick run through the first test scenario under the debugger points to the (first) obvious problem:  the test code is not setting `mock_requests_post.return_value.status_code = 200`, so the CUT decides that the request has failed and returns _early_, without doing anything which we would notice as a failure.

So, this is a little tricky to test:  we're trying to make sure that certain things _don't_ happen, while assuming that everything else _does_.  What I think we should do is set up the context so that it will set the priority value and then make sure that it doesn't set the story point value:  if it sets the story point value (or dies trying), then the test failed; if it _doesn't_ set the priority value, then the test failed, as well.

I think that's probably a workable approach, but I have to quit for the day.

webbnh (approved):
@istein1, it looks good to me, now!

Thanks for your patience and perseverance.  🙂

Bala-Sakabattula:
Deployed and merged please check the functionality

PR #454, opened 2026-04-13
comment format payload limit fixes

Bala-Sakabattula:
Changed the code if the comment payload limit exceeds 32767 characters than truncating the comment payload and adding the link of github issue url at the end.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -256,42 +256,46 @@ def _comment_format(comment):
     )
 
 
-def _truncate_jira_text(body: str, upstream_issue_url: Optional[str] = None) -> str:
+def _truncate_jira_text(
+    body: str,
+    upstream_issue_url: Optional[str] = None,
+    max_chars: Optional[int] = None,
+) -> str:
     """
-    Ensure ``body`` fits Jira's comment size limit.
+    Ensure ``body`` fits within *max_chars* (default: comment limit).
 
     If truncated, a notice is prepended and a truncation marker appended.  When
     an upstream URL is available it is included — *unless* the link is so
     long that it would eat into the minimum body budget, in which case the link
     is dropped entirely.
     """
-    if len(body) <= JIRA_TEXT_BODY_MAX_CHARS:
+    if max_chars is None:
+        max_chars = JIRA_TEXT_BODY_MAX_CHARS
+
+    if len(body) <= max_chars:
         return body
 
     log.info(
-        "Truncating Jira comment body from %d to max %d characters",
+        "Truncating Jira text body from %d to max %d characters",
         len(body),
-        JIRA_TEXT_BODY_MAX_CHARS,
+        max_chars,
     )
 
     head = "{warning}*(Truncated.)*{warning}\n"
-    tail = "\n\n{warning}*....*{warning}"
+    tail = "\n\n{warning}*....*{warning}\n"
 
     link_block = ""
     if upstream_issue_url:
-        link_block = f"[See More|{upstream_issue_url}]\n\n"
+        link_block = f"[See More|{upstream_issue_url}]\n"
 
-    fixed_overhead = len(head) + len(tail)
+    fixed_overhead = len(head) + len(tail) + 1  # +1 for the \n before body
     link_overhead = 2 * len(link_block)
-    if (
-        fixed_overhead + link_overhead + JIRA_TEXT_BODY_MIN_CHARS
-        > JIRA_TEXT_BODY_MAX_CHARS
-    ):
+    if fixed_overhead + link_overhead + JIRA_TEXT_BODY_MIN_CHARS > max_chars:
         link_block = ""
         link_overhead = 0
 
-    core_len = JIRA_TEXT_BODY_MAX_CHARS - fixed_overhead - link_overhead
-    return head + link_block + body[:core_len] + tail + link_block
+    core_len = max_chars - fixed_overhead - link_overhead
+    return head + link_block + "\n" + body[:core_len] + tail + link_block
 
 
 def _comment_format_legacy(comment):
@@ -516,9 +520,7 @@ def check_comments_for_duplicate(client, result, username):
     return None
 
 
-def _find_comment_in_jira(
-    comment, j_comments, upstream_issue_url: Optional[str] = None
-):
+def _find_comment_in_jira(comment, j_comments, issue_url: Optional[str] = None):
     """
     Helper function to filter out comments that are matching.
 
@@ -533,9 +535,7 @@ def _find_comment_in_jira(
         # touch the comment; return the item as is.
         return comment
 
-    formatted_comment = _truncate_jira_text(
-        _comment_format(comment), upstream_issue_url
-    )
+    formatted_comment = _truncate_jira_text(_comment_format(comment), issue_url)
     legacy_formatted_comment = _comment_format_legacy(comment)
     for item in j_comments:
         if item.raw["body"] == legacy_formatted_comment:
@@ -1364,23 +1364,41 @@ def _update_tags(updates, existing, issue):
 
 
 def _build_description(issue):
-    # Build the description of the JIRA issue
+    # Build the description of the JIRA issue.
+    #
+    # Truncate issue.content *before* wrapping it with {quote} and the
+    # metadata prefix so the closing {quote} and decoration are never lost.
     issue_updates = issue.downstream.get("issue_updates", [])
-    description = ""
-    if "description" in issue_updates:
-        description = f"Upstream description: {{quote}}{issue.content}{{quote}}"
 
-    if issue.status and any("transition" in item for item in issue_updates):
-        # Add the upstream issue status to the top of the description
-        formatted_status = "Upstream issue status: " + issue.status
-        description = formatted_status + "\n" + description
-
+    # Build the prefix lines that will appear above the quoted content.
+    prefix_parts = []
     if issue.reporter:
-        # Add to the description
-        prefix = f"[{issue.id}] Upstream Reporter: {issue.reporter['fullname']}\n"
-        description = prefix + description
+        prefix_parts.append(
+            f"[{issue.id}] Upstream Reporter: {issue.reporter['fullname']}"
+        )
+    if issue.status and any("transition" in item for item in issue_updates):
+        prefix_parts.append("Upstream issue status: " + issue.status)
 
-    return _truncate_jira_text(description, getattr(issue, "url", None))
+    prefix = "\n".join(prefix_parts) + "\n" if prefix_parts else ""
+    suffix = ""
+    if "url" in issue_updates:
+        suffix = f"\nUpstream URL: {issue.url}"
+    if "description" in issue_updates:
+        quote_wrapper = "Upstream description: {quote}{quote}"
+        decoration_len = len(prefix) + len(quote_wrapper)
+        budget = JIRA_TEXT_BODY_MAX_CHARS - decoration_len
+        description = prefix
+        content = issue.content or ""
+        if len(content) > budget:
+            upstream_url = getattr(issue, "url", None)
+            content = _truncate_jira_text(content, upstream_url, max_chars=budget)
+            description += f"Upstream description: {{quote}}{content}{{quote}}"
+        else:
+            description += f"Upstream description: {{quote}}{content}{{quote}}" + suffix
+    else:
+        description = prefix + suffix
+
+    return description
 
 
 def _update_description(existing, issue):
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -46,6 +46,9 @@ load_dotenv()
 # This is used to ensure legacy comments are not touched
 UPDATE_DATE = datetime(2019, 7, 9, 18, 18, 36, 480291, tzinfo=timezone.utc)
 
+# Jira REST API rejects comment bodies longer than this (characters).
+JIRA_COMMENT_BODY_MAX_CHARS = 32767
+
 log = logging.getLogger("sync2jira")
 logging.getLogger("snowflake.connector").setLevel(logging.WARNING)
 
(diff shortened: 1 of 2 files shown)


sync2jira/downstream_issue.py
```diff
@@ -250,6 +253,52 @@ def _comment_format(comment):
...
+    )
+    tail = (
+        "\n\n{warning}*(Comment truncated"
+        + (
+            " — see GitHub link above for the full thread.)*"
+            if upstream_issue_url
+            else " — full thread not linked.)*"
+        )
+        + "{warning}"
+    )
```
webbnh, line 290: I don't think it's worth the effort just to be able to say, "go look at the top" or "don't bother looking at the top".

What we need to indicate is that the comment got truncated by the tooling.  If you want to supply more than that, I would use the value of `link_block` again (so that the user doesn't have to scroll back to the top of the comment in order to find the hyperlink).  If there is no hyperlink, I don't think we need to point that out here.

sync2jira/downstream_issue.py
```diff
@@ -250,6 +253,52 @@ def _comment_format(comment):
...
+
+    budget = JIRA_COMMENT_BODY_MAX_CHARS - len(head) - len(tail)
+    if budget < 1:
+        head = "{warning}*(Truncated.)*{warning}\n"
+        tail = "\n{warning}*(…)*{warning}"
+        budget = JIRA_COMMENT_BODY_MAX_CHARS - len(head) - len(tail)
+
+    truncated_core = body[: max(budget, 0)]
+    return head + truncated_core + tail
```
webbnh, line 299: This is excellent defensive coding, but I don't really like how it turned out.

Assuming that we have a reasonable budget (which we can confirm with a unit test), we don't need to fall back to quite such brief head and tail.  The only reason we would be over budget is if we received a _crazy_ long URL value, so that's really the only thing we would need to drop.

So, I propose that we don't add the `link_block` to the header, but instead we carry it through separately.  That would allow us to omit it easily if it is outrageous.

This would lead to code like,
```suggestion
    if len(head) + len(tail) + 2 * len(link_block) > JIRA_COMMENT_BODY_MAX_CHARS:
        link_block = ""
    core_len = JIRA_COMMENT_BODY_MAX_CHARS - len(head) - len(tail) - 2 * len(link_block)
    return head + link_block + body[: core_len] + tail + link_block
```

With this approach, `head` and `tail` are basically constants which could be defined at the module level (a module level definition of `head` could still depend on the value of `JIRA_COMMENT_BODY_MAX_CHARS` if you define them in order...but, it should have a "module-level name"...), and you could include a module-level assertion that the length of `head` plus the length of `tail` is less than `JIRA_COMMENT_BODY_MAX_CHARS` (or you could do it in a unit test).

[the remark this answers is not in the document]
webbnh: This is a valid and appropriate point, but I don't think it is practical to address it.  (It would likely be a fun project, but I don't think it would be worth our time and effort.)

sync2jira/downstream_issue.py
```diff
@@ -46,6 +46,9 @@
 # This is used to ensure legacy comments are not touched
 UPDATE_DATE = datetime(2019, 7, 9, 18, 18, 36, 480291, tzinfo=timezone.utc)
 
+# Jira REST API rejects comment bodies longer than this (characters).
+JIRA_COMMENT_BODY_MAX_CHARS = 32767
```
webbnh, line 50: We probably shouldn't set this to the literal maximum.  We can afford to leave ourself a little bit of headroom (e.g., for [silliness like this](https://github.com/Bala-Sakabattula/Sync2Jira/blob/33fce4b552af8b06dc3ff4f91aae094d6f1376d7/sync2jira/downstream_issue.py#L626)).

So, set it to `32760` or `32750` or `32700` or `32000`....

sync2jira/downstream_issue.py
webbnh, line 1392: We should probably be applying `_truncate_jira_comment_body()` somewhere in here, too (like probably at line 1376).

Note that this function wraps a bunch of stuff around the description body, so we'll need to leave room in (or outside of...) our "budget" for that.

tests/test_downstream_issue.py
```diff
@@ -1730,6 +1730,41 @@ def test_jira_user_display_label(self):
...
+        """
+        When the issue URL is huge, head+link can leave no room for body; cover
+        the minimal head/tail fallback (budget < 1 branch).
+        """
+        huge_url = "https://github.com/o/r/issues/1" + ("x" * 40000)
+        body = "D" * (d.JIRA_COMMENT_BODY_MAX_CHARS + 200)
+        out = d._truncate_jira_comment_body(body, huge_url)
+        self.assertLessEqual(len(out), d.JIRA_COMMENT_BODY_MAX_CHARS)
+        self.assertIn("*(Truncated.)*", out)
+        self.assertIn("*(…)*", out)
```
webbnh, line 1766: It seems like we should confirm that a certain number of `D`'s are in there, too.  😇 

(A similar comment applies to each of the previous two tests, as well.)

webbnh (commented):
This is a great first pass, but there are a few things that you should attend to.
- I assume that the PR _description_ body is subject to the same constraints as the comments, so I think we should be prepared to truncate that, too.
- Handling a crazy-long URL is an excellent piece of defensive coding, but I think we can be more strategic and graceful in how to do it.
- I think we should leave some "head room" in (or outside) the budget for add-ons that we're not anticipating (I called out a couple of them) so that we don't fail on edge cases.
- And, I think we have a couple more things that the unit tests should assert.

sync2jira/downstream_issue.py
webbnh: > ```
> if len(head) + len(tail) + 2 * len(link_block) > JIRA_COMMENT_BODY_MAX_CHARS:
> ```

As I think about it, we should probably adjust this to allow for at least _some_ of the comment body.  That is, even if we can fit the head, tail and hyperlinks, if we cannot fit, say, 1kb of the comment, then we should omit the hyperlink.  (That is, I think the most important part of the payload is the first part of the comment body; if we cannot fit the whole comment, then the second most important part is the hyperlink; if we can fit the hyperlink (twice), then include as much of the body as we reasonably can; however, if we can't fit everything, then truncate the comment up until it's almost but not quite gone, and then throw away the hyperlink(s).)

So, the expression should be something more like,
```python
    if len(head) + len(tail) + 2 * len(link_block) + JIRA_COMMENT_BODY_MIN_CHARS > JIRA_COMMENT_BODY_MAX_CHARS
```
(but Black is going to make a mess of that, so we'll probably want to instantiate a local variable or two).

Bala-Sakabattula: Why two link blocks are required ??? Like we can reduce it to one and add comment body instead right ???

sync2jira/downstream_issue.py
```diff
@@ -250,6 +256,44 @@ def _comment_format(comment):
...
+        len(body),
+        JIRA_TEXT_BODY_MAX_CHARS,
+    )
+
+    head = "{warning}*(Truncated.)*{warning}\n"
+    tail = "\n\n{warning}*....*{warning}"
+
+    link_block = ""
+    if upstream_issue_url:
+        link_block = f"[See More|{upstream_issue_url}]\n\n"
```
webbnh, line 282: I assume that the plan/desire is to have the following:
- the head on one line
- the link on the next line
- a blank line
- the body
- a blank line
- the tail line
- the link
- (end of comment; no trailing blank line)

The current distribution of `\n` characters won't produce that very well.

I think you should have exactly one newline at the end of each of the head, tail, and link (but keep the double-newline at the beginning of the tail), and then in the `return` expression add in the newline before the body explicitly.  That way, we won't end up with a spurious trailing newline from the second link block, and we won't lose the first blank line if the link is omitted.

(Although, if that one, extra newline isn't included in the link field, you'll need to account for it in `core_len` or it will have to come out of the "outside" space excluded from `JIRA_TEXT_BODY_MAX_CHARS`.  😉)

sync2jira/downstream_issue.py
```diff
@@ -486,7 +533,9 @@ def _find_comment_in_jira(comment, j_comments):
         # touch the comment; return the item as is.
         return comment
 
-    formatted_comment = _comment_format(comment)
+    formatted_comment = _truncate_jira_text(
+        _comment_format(comment), upstream_issue_url
+    )
```
webbnh, line 538: Black won't leave that on one line, eh?  😞

Maybe we should shorten `upstream_issue_url` to `issue_url` or similar....  😇  (That might allow lines 519-521 to be combined in to a single line, as well.  🙂)

sync2jira/downstream_issue.py
```diff
@@ -1330,11 +1380,7 @@ def _build_description(issue):
         prefix = f"[{issue.id}] Upstream Reporter: {issue.reporter['fullname']}\n"
         description = prefix + description
 
-    # Add the url if requested
-    if "url" in issue_updates:
-        description = description + f"\nUpstream URL: {issue.url}"
-
-    return description
+    return _truncate_jira_text(description, getattr(issue, "url", None))
```
webbnh, line 1384: I don't think that these two pieces of code are necessarily equivalent:  in the new code, the URL will be added (unconditionally) if the description (etc.) is more than 32Kb; in the old code the URL will be added any time that it is requested by the user configuration.

So, you've removed the code which would have added the URL to the description, presumably to prevent it from showing up twice in a row, because `_truncate_jira_text()` will also add it; however, it won't show up twice, because `_truncate_jira_text()` will only add it if it is truncating the text, in which case the added URL will be the first thing removed, because it was added at the end of the description.

But, of even more concern is that fact that you're truncating the description _after_ it's been wrapped with all of the stuff from (old) lines 1320-1335, in particular, the truncation will remove the closing `{quote}` directive added at (old) line 1321.

If we're going to truncate the message, I really think we should do it before line 1321, so that all of our "decoration" comes through properly.  (If we know that we're going to add the URL as part of the decoration, then we can omit it in the call to `_truncate_jira_text()` so that it doesn't appear twice.)  However, since we won't account for all of the decoration when we're inside `_truncate_jira_text()`, we'll want to decrease the value of `JIRA_TEXT_BODY_MAX_CHARS` to leave room for it.

tests/test_downstream_issue.py
```diff
@@ -1730,6 +1749,45 @@ def test_jira_user_display_label(self):
...
+        self.assertEqual(out.count(link), 2)
+        self.assertTrue(out.startswith("{warning}"))
+        self.assertGreaterEqual(out.count("B"), d.JIRA_TEXT_BODY_MIN_CHARS)
+
+    def test_truncate_jira_text_long_without_url(self):
+        body = "C" * (d.JIRA_TEXT_BODY_MAX_CHARS + 100)
+        out = d._truncate_jira_text(body, None)
+        self.assertLessEqual(len(out), d.JIRA_TEXT_BODY_MAX_CHARS)
+        self.assertIn("....", out)
+        self.assertGreaterEqual(out.count("C"), d.JIRA_TEXT_BODY_MIN_CHARS)
```
webbnh, line 1776: It seems like we should have the same (sub)set of assertions in each of the (non-short) tests, here:
- Check for "Truncated"
- Check for the expected presence or absence of the link blocks
- Check for the trailing elipsis
- Check for the presence of the minimum body

Those should hold in each case as we vary in the inputs.  And, the fact that we checked for them in other scenarios doesn't mean that we _shouldn't_ check for them in subsequent scenarios.  (Each test should be self-contained and not count on other tests for completeness...because the other tests might be modified or removed...and because with different inputs, the outputs might be different.)

sync2jira/downstream_issue.py
webbnh: Think about what your user experience would be like viewing a Jira ticket with at least one comment containing 32Kb (to get an idea, look [here](https://github.com/quarkusio/search.quarkus.io/issues/130#issuecomment-1913376614)) -- after you've been scrolling and scrolling you get to the bottom of the comment and find that it was truncated:  would you like to have a link to click on right there, or do you want to be forced to scroll back to the top of the comment to find it?  (And, after you've been through that once, the next time you'll want to click on the link at the top of the comment!)

So, yeah, I think it would be good to have the links in both places.  And, in general, I think the space taken up by the URL will be very small (you've seen the GH URLs...).

Once we decide to truncate a comment, the fraction of the comment text that we have to give up in order to fit the URL is tiny (less than 0.5%).  And, if we have to truncate it, the odds of losing something important that wouldn't be omitted if we skipped the link are just as tiny.

webbnh (commented):
Getting close, but there are a couple of new things you probably want to fix, and there are a couple of other things to consider.

sync2jira/downstream_issue.py
Bala-Sakabattula: if we do before than we have to do decrease JIRA_TEXT_BODY_MAX_CHARS or for description we have to pass new value as if we decrease then comment will also be affected.

sync2jira/downstream_issue.py
Bala-Sakabattula: I already added the link block at the top and end .

webbnh: (My apologies -- I forgot to add the hyperlink to the Issue with the huge comment; that's fixed, now.)

sync2jira/downstream_issue.py
```diff
@@ -1364,23 +1364,41 @@ def _update_tags(updates, existing, issue):
...
-        # Add the upstream issue status to the top of the description
-        formatted_status = "Upstream issue status: " + issue.status
-        description = formatted_status + "\n" + description
+        prefix_parts.append("Upstream issue status: " + issue.status)
 
-    if issue.reporter:
-        # Add to the description
-        prefix = f"[{issue.id}] Upstream Reporter: {issue.reporter['fullname']}\n"
-        description = prefix + description
+    prefix = "\n".join(prefix_parts) + "\n" if prefix_parts else ""
```
webbnh, line 1382: Building a list of lines and then joining them together is fine, but, instead of appending parts to the list, why not just append them to the string itself (which is basically what `join()` is ultimately going to do)?  It's simpler, and you don't have to do the join operation.

webbnh (approved):
Looks good to me.

I have one thought, but you don't need to address it unless you want to.  😇

sync2jira/downstream_issue.py
Bala-Sakabattula: I did list and join as it is easy to handle  "\n".join(parts) is that you never get a double \n if one of the middle parts is skipped. In normal string I need to take care of that..

webbnh: If you include the newline in the "part", then it is present when when you add it an omitted when you skip it (and so you don't need the ternary expression that you currently have after the `join()` call).

But, if you like the current approach better, I'm fine with it.

PR #457, opened 2026-04-19
LDAP IPA access

Bala-Sakabattula:
Removed anoymous access of LDAP made the service account mandatory to access the LDAP service and also removed default hardcoded server name and bind_dn we can pass it through the config which is better and secure.

### DIFF
--- a/Rover_Lookup/tests/test_lookup.py
+++ b/Rover_Lookup/tests/test_lookup.py
@@ -50,13 +50,16 @@ class TestGithubUsernameToEmails:
     @patch("Rover_Lookup.lookup.Connection")
     def test_successful_lookup_single_record(self, mock_connection_class, caplog):
         """Test successful LDAP lookup with single record containing multiple emails."""
+        # Mock the LDAP server connection
         mock_conn = Mock()
         mock_connection_class.return_value = mock_conn
         mock_conn.search.return_value = True
 
+        # Mock the LDAP entry
         mock_entry = Mock()
         mock_entry.entry_dn = "cn=testuser,cn=users,cn=accounts,dc=redhat,dc=com"
 
+        # Mock email attributes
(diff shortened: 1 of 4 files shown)


Rover_Lookup/tests/test_lookup.py
```diff
@@ -53,28 +66,34 @@ def test_successful_lookup_single_record(self, mock_connection_class, caplog):
         mock_entry.mail = mock_mail
 
         mock_preferred_alias = Mock()
-        # Duplicate value to test deduplication
         mock_preferred_alias.value = "user@redhat.com"
```
webbnh, line 57: We probably shouldn't remove this comment...otherwise, in the future, we might wonder why the test uses the same email address here and at (new) line 61.

Rover_Lookup/tests/test_lookup.py
```diff
@@ -144,11 +160,9 @@ def test_records_with_no_emails(self, mock_connection_class):
...
 
-        # Mock entry with no email attributes
         mock_entry = Mock()
-        mock_entry.entry_dn = "cn=testuser,ou=users,dc=redhat,dc=com"
+        mock_entry.entry_dn = "cn=testuser,cn=users,cn=accounts,dc=redhat,dc=com"
 
-        # Mock attributes that don't exist or are empty
         mock_entry.rhatPrimaryMail = Mock()
         mock_entry.rhatPrimaryMail.value = None
         mock_entry.mail = Mock()
```
webbnh, line 154: We might want to retain these line comments, too, since they explain/highlight stuff that we're _omitting_.

webbnh (dismissed):
Looks good, but there are a couple of code comments that I think you should consider retaining.

webbnh (approved):
👍

PR #460, opened 2026-04-25
Add branch/type filters to transitions and dict-based fixVersion mapping

istein1:
## Summary
- Add `branches` (glob) and `issue_types` filters to PR transition configs, so transitions only fire for matching branches and Jira issue types
- Add `issue_types` filter support to issue transition configs
- Support dict-based `fixVersion` mapping (in addition to existing string template format)
- Add `base_branch` field to PR intermediary, extracted from GitHub PR payload
- Add comprehensive tests for all new filtering and mapping logic

## Test plan
- [x] Table-driven tests for `_update_transition` with issue_types filtering (7 scenarios)
- [x] Tests for PR branch filter match/no-match
- [x] Tests for PR issue type filter match/no-match
- [x] Test for multiple transition entries selecting the correct match
- [x] Table-driven tests for `map_fixVersion` (string template + dict lookup)
- [x] End-to-end test for dict-based fixVersion mapping through `Issue.from_github`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/config-file.rst
+++ b/docs/source/config-file.rst
@@ -167,8 +167,10 @@ The config file is made up of multiple parts
         * Sync description
     * :code:`'title'`
         * Sync title
-    * :code:`{'transition': True/'CUSTOM_TRANSITION'}`
-        * Sync status (open/closed), Sync only status/Attempt to transition JIRA ticket to CUSTOM_TRANSITION on upstream closure
+    * :code:`{'transition': 'CUSTOM_TRANSITION'}`
+        * Sync status (open/closed). Attempt to transition JIRA ticket to CUSTOM_TRANSITION on upstream closure.
+    * :code:`{'transition': 'CUSTOM_TRANSITION', 'issue_types': ['Bug', 'Story']}`
+        * Same as above, but the transition only fires when the downstream JIRA issue type is in the list.
     * :code:`{'on_close': {'apply_labels': ['label', ...]}}`
         * When the upstream issue is closed, apply additional labels on the corresponding Jira ticket.
     * :code:`github_markdown`
@@ -193,16 +195,23 @@ The config file is made up of multiple parts
   downstream components. You can add the following to the :code:`pr_updates` array:
 
     * :code:`{'merge_transition': 'CUSTOM_TRANSITION'}`
-        * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge
+        * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge.
     * :code:`{'link_transition': 'CUSTOM_TRANSITION'}`
-        * Sync when upstream PR gets linked. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream link
+        * Sync when upstream PR gets linked. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream link.
+    * :code:`'branches': ['release-*', 'main']`
+        * Optional filter (glob patterns). The transition only fires when the PR's target branch matches one of the patterns.
+    * :code:`'issue_types': ['Bug', 'Story']`
+        * Optional filter. The transition only fires when the downstream JIRA issue type is in the list.
 
 * You can add the following to the mapping array. This array will map an upstream field to the downstream counterpart
   with XXX replaced.
 
     * :code:`{'fixVersion': 'Test XXX'}`
-        * Maps upstream milestone (suppose it's called 'milestone') to downstream fixVersion with a mapping (for our
-          example it would be 'Test milestone')
+        * String template format. Maps upstream milestone (suppose it's called 'milestone') to downstream fixVersion
+          with a mapping (for our example it would be 'Test milestone').
+    * :code:`{'fixVersion': {'0.9.0': 'Product 8.1', '1.0.0': 'Product 9.0'}}`
+        * Dict lookup format. Maps specific upstream milestones to specific downstream fixVersions.
+          Milestones not present in the dict are left unchanged.
 
 * It is strongly encouraged for teams to use the :code:`owner` field. If configured, owners will be alerted if Sync2Jira
   finds duplicate downstream issues. Further the owner will be used as a default in case the program is unable to find a
--- a/docs/source/config-file.rst
+++ b/docs/source/config-file.rst
@@ -167,10 +167,10 @@ The config file is made up of multiple parts
         * Sync description
     * :code:`'title'`
         * Sync title
-    * :code:`{'transition': 'CUSTOM_TRANSITION'}`
-        * Sync status (open/closed). Attempt to transition JIRA ticket to CUSTOM_TRANSITION on upstream closure.
     * :code:`{'transition': 'CUSTOM_TRANSITION', 'issue_types': ['Bug', 'Story']}`
-        * Same as above, but the transition only fires when the downstream JIRA issue type is in the list.
+        * Sync status (open/closed). Attempt to transition JIRA ticket to CUSTOM_TRANSITION on upstream closure.
+        * ``issue_types`` is optional and may be omitted. When present, the transition only fires if the
+          downstream JIRA issue type is in the list.
     * :code:`{'on_close': {'apply_labels': ['label', ...]}}`
         * When the upstream issue is closed, apply additional labels on the corresponding Jira ticket.
     * :code:`github_markdown`
@@ -194,17 +194,16 @@ The config file is made up of multiple parts
 * You can add your projects here. The 'project' field is associated with downstream JIRA projects, and 'component' with
   downstream components. You can add the following to the :code:`pr_updates` array:
 
-    * :code:`{'merge_transition': 'CUSTOM_TRANSITION'}`
+    * :code:`{'merge_transition': 'MODIFIED', 'branches': ['release-*', 'main'], 'issue_types': ['Bug', 'Story']}`
         * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge.
+        * ``branches`` and ``issue_types`` are optional and either may be omitted. ``branches`` accepts glob patterns
+          and restricts the transition to PRs whose target branch matches. ``issue_types`` restricts it to matching
+          downstream JIRA issue types.
     * :code:`{'link_transition': 'CUSTOM_TRANSITION'}`
         * Sync when upstream PR gets linked. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream link.
-    * :code:`{'merge_transition': 'MODIFIED', 'branches': ['release-*', 'main'], 'issue_types': ['Bug', 'Story']}`
-        * Optional filters added to the same dict as the transition. ``branches`` accepts glob patterns and restricts
-          the transition to PRs whose target branch matches. ``issue_types`` restricts it to matching downstream JIRA
-          issue types. Both filters are optional and can be used independently or together.
 
 * You can add the following to the mapping array. This array will map an upstream field to the downstream counterpart
-  with XXX replaced.
+  using either a template or a mapping table.
 
     * :code:`{'fixVersion': 'Test XXX'}`
         * String template format. Maps upstream milestone (suppose it's called 'milestone') to downstream fixVersion
--- a/docs/source/config-file.rst
+++ b/docs/source/config-file.rst
@@ -198,10 +198,10 @@ The config file is made up of multiple parts
         * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge.
     * :code:`{'link_transition': 'CUSTOM_TRANSITION'}`
         * Sync when upstream PR gets linked. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream link.
-    * :code:`'branches': ['release-*', 'main']`
-        * Optional filter (glob patterns). The transition only fires when the PR's target branch matches one of the patterns.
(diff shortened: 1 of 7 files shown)


sync2jira/downstream_pr.py
```diff
@@ -141,25 +142,76 @@ def update_jira_issue(existing, pr, client):
...
+        log.info(f"Updated {transition_type} for issue {pr.title}")
+        return
 
-    log.info(f"Updated {transition_type} for issue {pr.title}")
+    log.info(
+        "No matching %s entry for PR %s (branch=%s)",
+        transition_type,
+        pr.title,
+        pr.base_branch,
+    )
```
webbnh: I think this is going to produce a log message for every PR in every project which _doesn't_ configure `"pr_updates"`...which is going to be unpleasantly noisy, since that's the default configuration.

We should skip the log if `pr.downstream.get("pr_updates")` is false-y.

sync2jira/intermediary.py
```diff
@@ -226,6 +228,11 @@ def from_github(cls, upstream, pr, suffix, config, action=None):
         elif suffix not in lifecycle:
             suffix = "open"
 
+        # Extract the target branch from the PR payload
+        base_branch = (
+            pr.get("base", {}).get("ref") if isinstance(pr.get("base"), dict) else None
+        )
```
webbnh, line 233: Given the `isinstance()` test, you can replace the first `get()` call with a dict reference (and then it all fits on one line).

```suggestion
        base_branch = pr["base"].get("ref") if isinstance(pr.get("base"), dict) else None
```

tests/test_downstream_issue.py
```diff
@@ -1291,6 +1291,111 @@ def test_update_transition_successful(self, mock_client):
...
+            (
+                "multiple entries selects matching",
+                [
+                    {"transition": "MODIFIED", "issue_types": ["Bug"]},
+                    {"transition": "DEV COMPLETE", "issue_types": ["Story", "Task"]},
+                ],
+                "Story",
+                "Closed",
+                True,
+            ),
```
webbnh: While it is definitely good to have at least one scenario with multiple entries, I don't think the test can tell (as currently structured, at least) which of the entries caused the match...so, while the comment is descriptive and supportive, it's more a description of the intent than the expression of a requirement.  (But, I don't have any ideas on how to make the comment or the test better.)

I'd like to suggest having a scenario with multiple entries in which _none_ of them match, but I'm not sure it would provide very much benefit, either.  (The same logic as in the earlier comment applies:  it would produce no discernible effect, but it is a valid behavior situation.)

sync2jira/downstream_issue.py
```diff
@@ -1070,30 +1070,45 @@ def _update_transition(client, existing, issue):
...
+                    jira_type,
+                    type_filters,
+                )
+                continue
+
+        if (
+            closed_status is not True
+            and issue.status == "Closed"
+            and existing.fields.status.name.upper() != closed_status.upper()
+        ):
```
webbnh, line 1115: @ralphbean , what is it supposed to mean when `closed_status` is `True` (i.e., when the config key `'issue_updates'` includes `{'transition': True}`)?  (As it happens, nobody appears to do this [except `release-engineering/Sync2Jira`](https://gitlab.cee.redhat.com/rbean/sync2jira-config/-/blob/main/sync2jira.yml?ref_type=heads#L961)....)

The [documentation](https://github.com/release-engineering/Sync2Jira/blob/c03ba5c9e56693062b6b547f34116d7543e91ade/docs/source/config-file.rst?plain=1#L170) is less than clear, and my interpretation of it doesn't match my interpretation of this code....  (Should that be "if `closed_status` _is_ `True` _**or**_ the existing status doesn't match"?)

`True` is an acceptable value, but `False` _is **not**_...which is weird!  And, if we're the only ones using this, now might be a good time to rescind the practice.  (I think that `True` is supposed to be equivalent to `"Closed"`.)

tests/test_downstream_issue.py
```diff
@@ -1291,6 +1291,111 @@ def test_update_transition_successful(self, mock_client):
...
+            # 7: Upstream issue not closed — no transition fires
+            (
+                "upstream not closed",
+                [{"transition": "CLOSED"}],
+                "Bug",
+                "Open",
+                False,
+            ),
```
webbnh: This is good, but we should probably test the other two parts of [the condition](https://github.com/istein1/Sync2Jira/blob/369e086175446f2734ac9f321fd1c18cd11556d5/sync2jira/downstream_issue.py#L1100-L1104).

tests/test_downstream_issue.py
```diff
@@ -1291,6 +1291,111 @@ def test_update_transition_successful(self, mock_client):
...
 
+    @mock.patch("jira.client.JIRA")
+    def test_update_transition_issue_types_filter(self, mock_client):
+        """
+        Table-driven test for _update_transition with issue_types filtering.
+        Each scenario is cumulative, adding one piece of state.
+        """
+        mock_client.transitions.return_value = [{"name": "CLOSED", "id": "1234"}]
+
+        scenarios = (
```
webbnh, line 1345: Contrary to the statement in line 1298, I don't think the scenarios _are_ cumulative -- each one reestablishes all of the pertinent state.  (There is no dependency in any one scenario on any of the previous scenarios.)

There a few missing test scenarios:
- `"issue_updates"` key absent from `issue.downstream` (this is tricky to implement in the current structure)
- `issue.downstream["issue_updates"]` is an empty list

Both of these produce no observable effect (other than the absence of an observable effect), which is not very compelling, but, since these are valid configurations, it would be good to ensure that they don't cause anything to fall over.

tests/test_downstream_pr.py
```diff
@@ -373,6 +374,118 @@ def test_update_transition(self, mock_d_issue):
...
+        self.mock_pr.base_branch = "release-0.9"
+        self.mock_pr.downstream = {
+            "pr_updates": [
+                {"merge_transition": "MODIFIED", "branches": ["release-*"]},
+            ]
+        }
+
+        d.update_transition(
+            mock_client, self.mock_existing, self.mock_pr, "merge_transition"
+        )
```
webbnh: I think it would be better to test `update_transition()` and `_matches_transition_filters()` as separate units (I count five scenarios for the former and six for the latter...if you test them together, there are still _nine_ scenarios, and the setup is more complicated).

Part of the purpose of implementing `_matches_transition_filters()` as a subroutine is to encapsulate and abstract its functionality away from `update_transition()`...and that means that they can and should be tested separately.

webbnh (changes requested):
Hi @istein1, this looks great.

However, I've got a few items for you.  I've got a concern about one log message which I think might get noisy; I have a question for Ralph; and I have a tiny coding suggestion.  The rest of my comments are on the tests.

Your initial coverage looks very good, but there are some cases which I think should have additional scenarios, even though they won't increase the coverage, because they represent valid use situations, so I think we should exercise them in the tests.  Also, I recommend covering your new `_matches_transition_filters()` function as a separate unit (and maybe with the table-driven structure, if that works).

Lastly, I think this change should include a documentation update, since you're adding a couple of new configuration features.  ([Here](https://github.com/release-engineering/Sync2Jira/blob/main/docs/source/config-file.rst), at the minimum.)

Thanks!

istein1:
Thank you @webbnh,
Really appreciate your thorough review.
I'll work on addressing all the mentioned points.

istein1:
@webbnh ,Thanks for the detailed feedback! 
I’ve pushed updates to address all the comments above. 
Could you take another look and let me know if these changes satisfy your requirements, 
or if there’s anything else I should polish?

docs/source/config-file.rst
```diff
@@ -193,16 +195,23 @@ The config file is made up of multiple parts
...
     * :code:`{'merge_transition': 'CUSTOM_TRANSITION'}`
-        * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge
+        * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge.
     * :code:`{'link_transition': 'CUSTOM_TRANSITION'}`
-        * Sync when upstream PR gets linked. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream link
+        * Sync when upstream PR gets linked. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream link.
+    * :code:`{'merge_transition': 'MODIFIED', 'branches': ['release-*', 'main'], 'issue_types': ['Bug', 'Story']}`
+        * Optional filters added to the same dict as the transition. ``branches`` accepts glob patterns and restricts
+          the transition to PRs whose target branch matches. ``issue_types`` restricts it to matching downstream JIRA
+          issue types. Both filters are optional and can be used independently or together.
```
webbnh: We probably shouldn't have the `'link_transition'` between the two `'merge_transition'` descriptions.

And, I would be inclined to have just one `'merge_transition'` entry, perhaps like
```markdown
    * :code:`{'merge_transition': 'MODIFIED', 'branches': ['release-*', 'main'], 'issue_types': ['Bug', 'Story']}`
        * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge.
        * ``branches`` and ``issue_types`` are optional and either may be omitted. ``branches`` accepts glob patterns
          and restricts the transition to PRs whose target branch matches. ``issue_types`` restricts it to matching
          downstream JIRA issue types.
```

If you concur, I would recommend the same for `issue_updates` (lines 170-173).

docs/source/config-file.rst
```diff
@@ -193,16 +195,23 @@ The config file is made up of multiple parts
...
   with XXX replaced.
 
     * :code:`{'fixVersion': 'Test XXX'}`
-        * Maps upstream milestone (suppose it's called 'milestone') to downstream fixVersion with a mapping (for our
-          example it would be 'Test milestone')
+        * String template format. Maps upstream milestone (suppose it's called 'milestone') to downstream fixVersion
+          with a mapping (for our example it would be 'Test milestone').
+    * :code:`{'fixVersion': {'0.9.0': 'Product 8.1', '1.0.0': 'Product 9.0'}}`
+        * Dict lookup format. Maps specific upstream milestones to specific downstream fixVersions.
+          Milestones not present in the dict are left unchanged.
```
webbnh, line 213: You should probably tweak lines 206-207 as well.  E.g.,
```markdown
* You can add the following to the mapping array. This array will map an upstream field to the downstream counterpart
  using either a template or a mapping table.
```

webbnh (commented):
The code looks good; I just have a couple of thoughts on the doc changes.

sync2jira/downstream_issue.py
ralphbean: Yes. Let's refactor this and kill use of `True`. I can't explain what its supposed to mean. :D

docs/source/config-file.rst
istein1: Thanks. fixed.

sync2jira/downstream_issue.py
istein1: Thanks, fixed

istein1:
Thanks for the review @webbnh ,
Fixed the Docs comments, Normalized legacy transition: True to "Closed" (confirmed by Ralph),
and updated test expectation for True normalization.

docs/source/config-file.rst
```diff
@@ -194,17 +194,16 @@ The config file is made up of multiple parts
 * You can add your projects here. The 'project' field is associated with downstream JIRA projects, and 'component' with
   downstream components. You can add the following to the :code:`pr_updates` array:
 
-    * :code:`{'merge_transition': 'CUSTOM_TRANSITION'}`
+    * :code:`{'merge_transition': 'MODIFIED', 'branches': ['release-*', 'main'], 'issue_types': ['Bug', 'Story']}`
         * Sync when upstream PR gets merged. Attempts to transition JIRA ticket to CUSTOM_TRANSITION on upstream merge.
```
webbnh, line 198: In line 197, should we replace `'MODIFIED'` with `'CUSTOM_TRANSITION'`?  Line 198 references `CUSTOM_TRANSITION` as do lines 170-171.

sync2jira/downstream_issue.py
```diff
@@ -1085,6 +1085,12 @@ def _update_transition(client, existing, issue):
 
         closed_status = entry["transition"]
 
+        # Normalize legacy True value to "Closed"
+        if closed_status is True:
+            closed_status = "Closed"
+        if not isinstance(closed_status, str):
+            continue
```
webbnh, line 1098: FWIW, while offering compatibility for existing uses of `True` is nice, I'll cast another vote for _killing_ it (i.e., making it ignored or an error).

Speaking of which, if `closed_status` is not a string, I think we're looking at a malformed configuration, and we should probably log a warning for that before continuing.

webbnh (commented):
Nearly there!  I have a question about the latest doc change, and a comment/thought about the `True` transition type, and a small request for a diagnostic.

docs/source/config-file.rst
istein1: Good catch! fixed in a703e5a. Using CUSTOM_TRANSITION as the placeholder for consistency with the rest of the doc.

sync2jira/downstream_issue.py
istein1: Added a log.warning for non-string transition values in a703e5a. 
Regarding killing True support, I'd prefer to keep the normalization for now to avoid breaking existing configs, but happy to revisit in a follow-up if you and Ralph feel strongly about it.

istein1:
@webbnh , Any further changes needed?

webbnh (approved):
/lgtm
/ok-to-test

istein1:
Hi @ralphbean and @Zyzyx ,
Would you please review / merge?

PR #462, opened 2026-04-28
pypandoc tex format fix

Bala-Sakabattula:
If the `pypandoc` `convert` function fails due to TeX formatting issues, wrap the call in a `try` block. In the `except` block, clean or sanitize the TeX content and then retry the conversion.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -1521,6 +1521,29 @@ def sync_with_jira(issue, config):
         retry = True
 
 
+def convert_content(content):
+    """Convert GitHub Flavored Markdown to Jira format via pypandoc.
+
+    Falls back to disabling TeX math extensions if the initial conversion
+    fails (e.g. content misinterpreted as LaTeX).  Returns the original
+    content unchanged if both attempts fail.
+    """
+    try:
+        content = pypandoc.convert_text(content, "jira", format="gfm")
+    except Exception as e:
+        log.warning("pypandoc conversion failed, retrying with TeX disabled: %s", e)
+        try:
+            sanitized = content.replace("\\", "\\\\").replace("$", "\\$")
+            content = pypandoc.convert_text(
+                sanitized,
+                "jira",
+                format="gfm-tex_math_dollars-tex_math_single_backslash",
+            )
+        except Exception as e2:
+            log.warning("pypandoc fallback also failed; using raw content: %s", e2)
+    return content
+
+
 def update_jira(client, config, issue):
     # Check the status of the JIRA client
     if not config["sync2jira"]["develop"] and not check_jira_status(client):
@@ -1533,25 +1556,7 @@ def update_jira(client, config, issue):
             and issue.content
             and "github_markdown" in issue.downstream["issue_updates"]
         ):
-            try:
-                issue.content = pypandoc.convert_text(
-                    issue.content, "jira", format="gfm"
-                )
-            except Exception as e:
-                log.warning(
-                    "pypandoc conversion failed, retrying with TeX disabled: %s", e
-                )
-                try:
-                    sanitized = issue.content.replace("\\", "\\\\").replace("$", r"\$")
-                    issue.content = pypandoc.convert_text(
-                        sanitized,
-                        "jira",
-                        format="gfm-tex_math_dollars-tex_math_single_backslash",
-                    )
-                except Exception as e2:
-                    log.warning(
-                        "pypandoc fallback also failed; using raw content: %s", e2
-                    )
+            issue.content = convert_content(issue.content)
     # First, check to see if we have a matching issue using the new method.
     # If we do, then bail out.  No sync needed.
     log.info("Looking for matching downstream issue via new method.")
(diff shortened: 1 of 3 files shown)


sync2jira/downstream_issue.py
```diff
@@ -1533,8 +1533,25 @@ def update_jira(client, config, issue):
...
+            try:
+                issue.content = pypandoc.convert_text(
+                    issue.content, "jira", format="gfm"
+                )
+            except Exception as e:
+                log.warning(
+                    "pypandoc conversion failed, retrying with TeX disabled: %s", e
+                )
+                try:
+                    sanitized = issue.content.replace("\\", "\\\\").replace("$", r"\$")
```
webbnh: Why are you using a raw string for the second `replace()`'ment but not for the first?  I suggest being consistent.  Either:
```suggestion
                try:
                    sanitized = issue.content.replace("\\", "\\\\").replace("$", "\\$")
```
or

```suggestion
                try:
                    sanitized = issue.content.replace("\\", r"\\").replace("$", r"\$")
```

tests/test_downstream_issue.py
```diff
@@ -1110,6 +1110,73 @@ def test_sync_with_jira_no_matching(
         )
         mock_existing_jira_issue_legacy.assert_not_called()
 
+    @mock.patch(PATH + "get_jira_client")
+    @mock.patch(PATH + "get_existing_jira_issue")
+    @mock.patch(PATH + "_update_jira_issue")
+    @mock.patch(PATH + "_create_jira_issue")
```
webbnh: If `get_existing_jira_issue()` returns `True`, then we don't need to mock `_create_jira_issue()`; and, if we set `config["sync2jira"]["testing"]`, then we don't need to mock `_update_jira_issue()` either.

Actually, the pandoc conversion code now takes up enough space that it would probably be worth refactoring into a subroutine.  This would make `update_jira()` more readable, and it would make testing much simpler, since the test here would require no mocks other than for pandoc.  So, I would recommend moving lines 1531-1554 of `sync2jira/downstream_issue.py` (the body of the `if issue.downstream.get("issue_updates")` conditional) into its own function.

tests/test_downstream_issue.py
```diff
@@ -1110,6 +1110,73 @@ def test_sync_with_jira_no_matching(
...
+        self,
+        mock_pypandoc,
+        mock_check_jira_status,
+        mock_client,
+        mock_create_jira_issue,
+        mock_update_jira_issue,
+        mock_existing_jira_issue,
+        mock_get_jira_client,
+    ):
+        """Test pypandoc conversion: normal success, TeX fallback, and double failure."""
```
webbnh: Admittedly, the scenario or intention is to test the pypandoc conversion, unit tests are generally targeted at a "unit" of code, and the docstring should name that unit.  At the moment, the unit is `sync_with_jira()`.

That said, `sync_with_jira()` seems like a much bigger unit than what we want.  If you don't move the conversion code to its own unit, I think you should retool this test to target `update_jira()` instead, which will make the mocking requirements smaller.

webbnh (commented):
The change looks good (with one nit), but I think the unit test needs another pass; and, that will be much easier if you refactor the conversion code into its own function.  (The unit test is probably fine, but it would be better if it targeted a smaller unit.)

sync2jira/downstream_issue.py
```diff
@@ -1521,6 +1521,29 @@ def sync_with_jira(issue, config):
         retry = True
 
 
+def convert_content(content):
+    """Convert GitHub Flavored Markdown to Jira format via pypandoc.
+
+    Falls back to disabling TeX math extensions if the initial conversion
+    fails (e.g. content misinterpreted as LaTeX).  Returns the original
+    content unchanged if both attempts fail.
+    """
```
webbnh: This is missing both Python 3 type hints as well as the docstring-based type descriptions.  (I'm not so much a fan of the latter, but we should have at least one of the two!)

sync2jira/downstream_issue.py
```diff
@@ -1533,8 +1556,7 @@ def update_jira(client, config, issue):
             and issue.content
             and "github_markdown" in issue.downstream["issue_updates"]
         ):
-            issue.content = pypandoc.convert_text(issue.content, "jira", format="gfm")
-
```
webbnh, line 1537: If you edit this module for some other reason (like my previous comment), it would be good to restore the blank line between this call and the subsequent block comment.

sync2jira/downstream_issue.py
```diff
@@ -1533,8 +1556,7 @@ def update_jira(client, config, issue):
...
+            issue.content = convert_content(issue.content)
```
webbnh, line 1563: Should we be making [the same change for PRs](https://github.com/Bala-Sakabattula/Sync2Jira/blob/fd72dfcf41863cd452b556b8ce1ca1e39f7f3941/sync2jira/downstream_pr.py#L269)?

(This is why I wish that Issues and PRs used common code!)

webbnh (commented):
The test looks great, now.  However, the CUT needs a tweak or two, and, I think we want the same treatment for PRs (which might have more of a test impact, because I think we actually _have unit tests_ for PRs 🥴).

webbnh (approved):
👍

PR #470, opened 2026-05-13
PR to issue full sync

Bala-Sakabattula:
**Summary**
Extends the existing Sync2Jira integration to support full synchronization between Pull Requests and Jira Issues by reusing the existing Issue update pipeline, eliminating code duplication and ensuring consistent behavior across both sync types using Delegation pattern.

**Problem**
Previously, downstream_pr.py had its own minimal update_jira_issue that only handled three operations: PR link comments, merge_transition, and link_transition. All other update capabilities available to Issues (comments, description, title, assignee, tags, GitHub project fields, on_close labels) were not available for PRs.

**Solution**
**Unified update pipeline via updates_key parameter**

The core change introduces an updates_key parameter (defaulting to "issue_updates") to the shared update functions in downstream_issue.py, allowing the same code path to serve both Issues ("issue_updates") and PRs ("pr_updates"):

- _update_jira_issue(existing, issue, client, config, updates_key="issue_updates") — master orchestrator
- _update_transition(client, existing, issue, updates_key="issue_updates")
- _build_description(issue, updates_key="issue_updates")
- _update_description(existing, issue, updates_key="issue_updates")

**PR-specific operations remain separate**

Functions with no Issue equivalent stay in downstream_pr.py:

format_comment / issue_link_exists / comment_exists — PR link announcement
_update_pr_transitions — merge_transition (PR merged) and link_transition (PR first linked)

PR update_jira_issue now delegates to the shared pipeline:
``` 
def update_jira_issue(existing, pr, client, config):
    _update_pr_transitions(client, existing, pr)        # PR-specific
    d_issue._update_jira_issue(                          # Shared pipeline
        existing, pr, client, config, updates_key="pr_updates"
    ) 
```
**Files changed**

- sync2jira/downstream_issue.py — Added updates_key parameter to _update_jira_issue, _update_transition, _build_description, _update_description; extracted maybe_convert_markdown; made _build_description handle reporter as dict or string; used hasattr/getattr for Issue-only fields
- sync2jira/downstream_pr.py — Restructured: PR-specific functions kept (format_comment, issue_link_exists, comment_exists, _update_pr_transitions, _update_pr_transition); update_jira_issue now delegates to shared pipeline; update_jira uses maybe_convert_markdown
- fedmsg.d/sync2jira.py — Updated example config to show both issue_updates and pr_updates with inline documentation
- tests/test_downstream_pr.py — Updated assertions for new update_jira_issue signature (added config param); added shared pipeline delegation assertions; renamed test_update_transition to test_update_pr_transition
- tests/test_downstream_issue.py — Updated assertions for _update_description and _update_transition to include updates_key kwarg

**Backward compatibility**

- All updates_key parameters default to "issue_updates" — existing Issue configs are unchanged
- Existing pr_updates configs with merge_transition/link_transition continue to work as before


fedmsg.d/sync2jira.py
```diff
@@ -65,24 +65,58 @@
...
+                        'upstream_id',      # Add comment with upstream issue link on create
+                        'url',              # Include upstream URL in description
+                        'github_project_fields',  # Sync storypoints & priority from GitHub Projects
+                        {'transition': 'Closed'},  # Transition Jira when upstream issue closes
+                        {'assignee': {'overwrite': False}},  # Sync assignee (don't overwrite existing)
+                        {'on_close': {'apply_labels': ['closed-upstream']}},  # Label on close
+                    ],
+
+                    # ----- PR synchronization -----
+                    # pr_updates uses the same options as issue_updates.
```
webbnh: The `'pr_updates'` list here doesn't include `'upstream_id'` or `'url'` which the `'issues_updates'` list includes...so is the comment here wrong, is the omission an oversight, or is the implementation incomplete?

A similar comment applies to `'github_project_fields'`.

sync2jira/downstream_issue.py
```diff
@@ -985,18 +985,22 @@ def _label_matching(jira_labels, issue_labels):
     return updated_labels
 
 
-def _update_jira_issue(existing, issue, client, config):
+def _update_jira_issue(existing, issue, client, config, updates_key="issue_updates"):
```
webbnh: Since this function is now being called from outside this module, it should be renamed to remove the leading underscore.

sync2jira/downstream_issue.py
```diff
@@ -985,18 +985,22 @@ def _label_matching(jira_labels, issue_labels):
...
     """
     Updates an existing JIRA issue (i.e., tags, assignee, comments, etc.).
 
+    Works for both Issue and PR intermediary objects.  The ``updates_key``
+    parameter selects which downstream config list to read
+    (``"issue_updates"`` or ``"pr_updates"``).
```
webbnh: If you're going to use back-quotes at all in a comment, they should not be doubled.  (Given that you're already using double-quotes, I would omit the back-quotes from line 994.)

sync2jira/downstream_issue.py
```diff
@@ -1008,7 +1012,6 @@ def _update_jira_issue(existing, issue, client, config):
 
     # Get fields representing project item fields in GitHub and Jira
     github_project_fields = issue.downstream.get("github_project_fields", {})
-    # Only synchronize comments for listings that op-in
```
webbnh, line 1011: Given the comments at (old) lines 1018, 1023, 1028, etc., we should probably just _correct_ this comment rather than removing it.  (I.e., replace `comments` with `GitHub Project fields`.)

fedmsg.d/sync2jira.py
```diff
@@ -65,24 +65,58 @@
...
-                                                    'priority': {'gh_field': 'Priority', 'options':
-                                                                 {'P0': 'Blocker', 'P1': 'Critical', 'P2': 'Major',
-                                                                  'P3': 'Minor', 'P4': 'Optional', 'P5': 'Trivial'}}},
-                                                 'sync': ['pullrequest', 'issue']},
+                'GITHUB_USERNAME/Demo_project': {
+                    'project': 'FACTORY',
+                    'component': 'gitbz',
+
+                    # ----- Issue synchronization -----
+                    'issue_updates': [
```
webbnh, line 73: There's no example usage for `'fix_version'`...should there be?

sync2jira/downstream_issue.py
```diff
@@ -1043,7 +1047,7 @@ def _update_jira_issue(existing, issue, client, config):
     # Only synchronize descriptions for listings that op-in
     if "description" in updates:
         log.info("Looking for new description")
-        _update_description(existing, issue)
+        _update_description(existing, issue, updates_key=updates_key)
```
webbnh: I don't think we need to specify the keyword argument name for updates_key, so I would omit it in these cases (e.g., at new line 1062, as well, and lots of other places):  it doesn't provide any additional context (i.e., the keyword is the same as the variable name), and we don't provide keywords for any of the other arguments, so it's just visual clutter.

fedmsg.d/sync2jira.py
```diff
@@ -65,24 +65,58 @@
...
+                    # ----- PR synchronization -----
+                    # pr_updates uses the same options as issue_updates.
+                    # Additionally supports merge_transition and link_transition.
+                    'pr_updates': [
+                        'comments',         # Sync GitHub PR comments to Jira
+                        'title',            # Sync PR title
+                        'description',      # Sync PR description/body
+                        'github_markdown',  # Convert GitHub Markdown to Jira wiki format
+                        {'merge_transition': 'Closed'},    # Transition Jira when PR is merged
+                        {'link_transition': 'In Progress'},  # Transition Jira when PR is first linked
```
webbnh, line 101: Does `{'transition': 'Closed'}` work as expected when specified for PRs?  (If so, what exactly do we expect it to do?)

sync2jira/downstream_issue.py
```diff
@@ -1066,20 +1070,21 @@ def _update_jira_issue(existing, issue, client, config):
...
+    :param str updates_key: Config key for the updates list
     :returns: Nothing
     """
     # If the user added a custom closed status, attempt to close the
     # downstream JIRA ticket
 
     # First get the closed status from the config file
-    t = filter(lambda d: "transition" in d, issue.downstream.get("issue_updates", []))
+    t = filter(lambda d: "transition" in d, issue.downstream.get(updates_key, []))
     closed_status = next(t)["transition"]
```
webbnh: Note that this filter will _remove_ the `{'merge_transition': 'Closed'}` and `{'link_transition': 'In Progress'}` items...is that expected?  (That is, I don't think this function will do very much if `updates_key` is `"pr_updates"`....)

sync2jira/downstream_issue.py
```diff
@@ -1369,19 +1374,23 @@ def _update_tags(updates, existing, issue):
...
-    if issue.reporter:
-        prefix_parts.append(
-            f"[{issue.id}] Upstream Reporter: {issue.reporter['fullname']}"
-        )
+    reporter = getattr(issue, "reporter", None)
+    if reporter:
+        if isinstance(reporter, dict):
+            fullname = reporter.get("fullname", "")
+        else:
+            fullname = str(reporter)
```
webbnh: Ugh -- this is unfortunate code -- we _should_ know whether `reporter` is a `dict` or not before we get here....  (I see this code or code like it elsewhere -- I think it stems from confusion between the upstream object and the "intermediary" objects -- I think in the latter the `reporter` is always a `str` while in the former it is always a `dict`.)  It would be nice if we could remove this conditional....

sync2jira/downstream_issue.py
```diff
@@ -1407,16 +1416,17 @@ def _build_description(issue):
...
 
-def _update_description(existing, issue):
+def _update_description(existing, issue, updates_key="issue_updates"):
     """
     Helper function to sync description between upstream issue and downstream JIRA issue.
 
     :param jira.resource.Issue existing: Existing JIRA issue
-    :param sync2jira.intermediary.Issue issue: Upstream issue
+    :param issue: Upstream Issue or PR
+    :param str updates_key: Config key for the updates list
```
webbnh, line 1447: The `issue` line is missing its type.  (In the current code, it is `sync2jira.intermediary.Issue`...was that removed on purpose?)

sync2jira/downstream_pr.py
```diff
@@ -127,7 +128,7 @@ def update_jira_issue(existing, pr, client):
     # Only synchronize link_transition for listings that op-in
     if any("merge_transition" in item for item in updates) and "merged" in pr.suffix:
         log.info("Looking for new merged_transition")
-        update_transition(client, existing, pr, "merge_transition")
+        _update_pr_transition(client, existing, pr, "merge_transition")
```
webbnh: A nit, but I would reverse the order of the tests in the conditional:  there is no point in iterating over `updates` if the PR wasn't `"merged"`.

Ditto for (new) lines 135-139.

webbnh (commented):
This looks great!  I have a few nits and other bits of polish for you.  (Also, I haven't reviewed the tests, yet.)

fedmsg.d/sync2jira.py
Bala-Sakabattula: It won't work it would return as this condition will fail for PR 
```
if (
        closed_status is not True
        and issue.status == "Closed"
        and existing.fields.status.name.upper() != closed_status.upper()
    ):
```
   For PR transition we have other function _update_pr_transitions() which we call before the shared pipeline.

sync2jira/downstream_issue.py
Bala-Sakabattula: yes this function will work for issue only for PR transitions we have other function  _update_pr_transitions() which we call before the shared pipeline.

fedmsg.d/sync2jira.py
Bala-Sakabattula: Apologies its an oversight !! let me add them ..

fedmsg.d/sync2jira.py
Bala-Sakabattula: yes !! added!!

sync2jira/downstream_issue.py
Bala-Sakabattula: but now its can be Upstream issue or PR so removed the type

sync2jira/downstream_issue.py
Bala-Sakabattula: okay changed the reporter to be string for issue or PR..

fedmsg.d/sync2jira.py
```diff
@@ -78,19 +78,23 @@
...
+                    # plus merge_transition and link_transition.
                     'pr_updates': [
                         'comments',         # Sync GitHub PR comments to Jira
                         'title',            # Sync PR title
                         'description',      # Sync PR description/body
                         'github_markdown',  # Convert GitHub Markdown to Jira wiki format
+                        'upstream_id',      # Add comment with upstream PR link on create
+                        'url',              # Include upstream URL in description
+                        'github_project_fields',  # Sync storypoints & priority from GitHub Projects
                         {'merge_transition': 'Closed'},    # Transition Jira when PR is merged
```
webbnh, line 100: Now there's no example usage for `'fix_version'` for PRs...should there be?

sync2jira/downstream_issue.py
```diff
@@ -985,18 +985,22 @@ def _label_matching(jira_labels, issue_labels):
...
     :returns: Nothing
     """
-    # Start with comments
-    # Only synchronize comments for listings that op-in
-    log.info("Updating information for upstream issue: %s", issue.url)
+    log.info("Updating information for upstream %s: %s", updates_key, issue.url)
 
     # Get a list of what the user wants to update for the upstream issue
     updates = issue.downstream.get("issue_updates", [])
```
webbnh: This looks questionable:  given that `issue` might be a PR, shouldn't we be referencing `updates_key` instead of `"issue_updates"`, here??

You should search the sources for `"issue_updates"` -- I think I found four more references in `sync2jira/downstream_pr.py` which are suspect, as well.

sync2jira/downstream_issue.py
webbnh: Unlike all of the other, previous "Only...that opt-in" conditionals, this one doesn't check to see if the `updates` contain an opt-in.  Checking seems like a good idea, but, if we're not going to check, perhaps we should amend the comment.

sync2jira/downstream_issue.py
webbnh: OK, so `_update_pr_transitions()` handles `{'merge_transition': 'Closed'}` and `{'link_transition': 'In Progress'}` items, so they _should_ be removed, here.

> I don't think this function will do very much if updates_key is "pr_updates"....

In fact, the caller will not call this function if the `{'transition': ...}` item isn't present, so I think it's all good.

fedmsg.d/sync2jira.py
webbnh: > For PR transition we have other function _update_pr_transitions() which we call before the shared pipeline.

True, but the PR flow _also_ eventually calls the Issue `_update_transitions()` function, which _will_ process the `{'transition': 'Closed'}` item if it is present.

> It won't work it would return as this condition will fail for PR

I'm not sure what you are thinking when you say, "It won't work".  (Does the code need fixing??)

I _think_ that that conditional can be true for a PR.  (E.g., I think that `closed_status` can be `"Closed"`, and I _think_ that PRs can have a status named `"Closed"` (i.e., if they are closed without being merged).)

So, I _**think**_ that this code is operable for PRs...the question is, does it "work" (i.e., does it do something reasonable and predictable)?

sync2jira/downstream_issue.py
webbnh: Instead of omitting the type, you should provide both types.  You can list them separated by 'or':
```python
    :param sync2jira.intermediary.Issue or sync2jira.intermediary.PR issue: Upstream Issue or PR
```

sync2jira/intermediary.py
```diff
@@ -122,7 +122,7 @@ def from_github(cls, upstream, issue, config):
             fixVersion=[issue["milestone"]],
             priority=issue.get("priority"),
             content=issue["body"] or "",
-            reporter=issue["user"],
+            reporter=issue["user"]["fullname"],
```
webbnh, line 125: Are we confident that the `"fullname"` field is always present and always has a useful value?  (E.g., what do we receive when the upstream user doesn't provide their name in their profile?)

fedmsg.d/sync2jira.py
webbnh: No need to apologize!  This is why we do review....  🙂

webbnh (commented):
The update looks good, but I found some more stuff to consider (and, I still haven't gotten to the tests... 🥴).

tests/test_downstream_issue.py
```diff
@@ -747,7 +747,7 @@ def test_assign_user_remove_all(self, mock_client):
         mock_client.search_users.assert_not_called()
 
     def common_test_create_jira_issue(
-        self, mock_attach_link, mock_client, mock_update_jira_issue
+        self, mock_attach_link, mock_client, mockupdate_jira_issue
```
webbnh: Looks like your global replace went too far.  🙂  Even though the function name no longer starts with an underscore, the `mock_` prefix should still end with one.  😉 

Ditto line 801, et al.

tests/test_downstream_issue.py
```diff
@@ -1104,12 +1104,50 @@ def test_sync_with_jira_no_matching(
...
+        issue.content = "**bold**"
+        d.maybe_convert_markdown(issue, "pr_updates")
+        mock_convert.assert_called_once_with("**bold**")
+        self.assertEqual(issue.content, "*bold*")
+
+        # Skips when github_markdown not in updates
+        mock_convert.reset_mock()
+        issue.downstream = {"issue_updates": ["description"]}
+        issue.content = "# Hello"
+        d.maybe_convert_markdown(issue)
```
webbnh, line 1146: You have a case which tests the default value of the second parameter and one which tests the value `"pr_updates"`, but you don't have a case which tests `"issue_updates"`.

And, I don't think there is a case where the `updates_key` is not found; nor is there a case which tests when the value of the `updates_key` is empty (each of which is a reasonable log configuration).

tests/test_downstream_issue.py
```diff
@@ -1104,12 +1104,50 @@ def test_sync_with_jira_no_matching(
...
 
+    @mock.patch(PATH + "convert_content")
+    def test_maybe_convert_markdown(self, mock_convert):
+        """maybe_convert_markdown: converts when opted-in, skips otherwise."""
+        mock_convert.return_value = "h1. Hello"
+
+        # Converts when source=github, content non-empty, github_markdown in updates
+        issue = MagicMock()
+        issue.downstream = {"issue_updates": ["description", "github_markdown"]}
+        issue.source = "github"
```
webbnh, line 1121: You don't have a case where the `issue.source` is not `"github"`.

tests/test_downstream_issue.py
```diff
@@ -1104,12 +1104,50 @@ def test_sync_with_jira_no_matching(
...
+        mockupdate_jira_issue.assert_not_called()
         mock_create_jira_issue.assert_called_with(
             mock_client, self.mock_issue, self.mock_config
         )
         mock_existing_jira_issue_legacy.assert_not_called()
 
+    @mock.patch(PATH + "convert_content")
+    def test_maybe_convert_markdown(self, mock_convert):
+        """maybe_convert_markdown: converts when opted-in, skips otherwise."""
+        mock_convert.return_value = "h1. Hello"
```
webbnh: Since we overwrite `mock_convert.return_value` at line 1129, this assignment (at line 1116) should be moved into the block of statements which starts at 1118.

tests/test_downstream_issue.py
```diff
@@ -1104,12 +1104,50 @@ def test_sync_with_jira_no_matching(
...
+
+        # Converts when source=github, content non-empty, github_markdown in updates
+        issue = MagicMock()
+        issue.downstream = {"issue_updates": ["description", "github_markdown"]}
+        issue.source = "github"
+        issue.content = "# Hello"
+        d.maybe_convert_markdown(issue)
+        mock_convert.assert_called_once_with("# Hello")
+        self.assertEqual(issue.content, "h1. Hello")
```
webbnh: I think that these two assertions are basically redundant.  I don't think that we really care whether `convert()` is called -- I _think_ all that we care about is that the `issue.content` ends up with the right value.  In this case, it gets that value from the (mock) `convert()` call, so the second assertion confirms the same thing that the first one does.  (I don't think we care about the input to `convert()`...we don't want the test to be too tightly coupled to the CUT.)

So, I think I would be inclined to omit all of the `mock_convert` assertions, and rely instead of inspecting the `issue.content` value (for all the test scenarios).  (Aside from simplifying things, this also relieves us of the need to reset the mock.)

tests/test_downstream_issue.py
```diff
@@ -1104,12 +1104,50 @@ def test_sync_with_jira_no_matching(
...
+        d.maybe_convert_markdown(issue)
+        mock_convert.assert_not_called()
+        self.assertEqual(issue.content, "# Hello")
+
+        # Skips when content is empty
+        mock_convert.reset_mock()
+        issue.downstream = {"issue_updates": ["github_markdown"]}
+        issue.content = ""
+        d.maybe_convert_markdown(issue)
+        mock_convert.assert_not_called()
```
webbnh: As I said above, I think our concern is about the value of `issue.content` after the call to `maybe_convert_markdown()`, so I would include an assertion on that (instead of the one on `mock_convert`).

tests/test_downstream_issue.py
```diff
@@ -1162,7 +1200,7 @@ def test_convert_content(self, mock_pypandoc):
     @mock.patch(PATH + "_update_assignee")
     @mock.patch(PATH + "_update_on_close")
     @mock.patch("jira.client.JIRA")
-    def test_update_jira_issue_closed(
+    def testupdate_jira_issue_closed(
```
webbnh: I think the underscore between `test` and `update` is critical -- I'm not sure that this function will be recognized as a test without it.

Ditto at line 1264, et al.

webbnh (commented):
Done with the tests...which look generally good, but there are a number of places where I think your global find-and-replace needs to be undone.

Also, I noted several missing test cases for `maybe_convert_markdown()` which you might consider adding.

And, I have a couple of other comments/suggestions.

One cool note:  it looks like [the Coveralls site](https://coveralls.io/builds/79455835) is working (at the moment, for this PR...) -- you can actually load the annotated source comparisons successfully!

sync2jira/downstream_issue.py
Bala-Sakabattula: yeah !! let me modify it ...

sync2jira/downstream_issue.py
Bala-Sakabattula: I update the comment as the opt in check is done internally inside the function
```
for item in updates:
        if isinstance(item, dict):
            if new_labels := item.get("on_close", {}).get("apply_labels", []):
                break
    else:
        return
```

sync2jira/intermediary.py
Bala-Sakabattula: I guess so!!  we have reformat_github_common function in upstream_issue and upstream_pr which have fallback code if  GitHub user doesn't provide their name (i.e., reporter.name is None or the string "None"), it falls back to item["user"]["login"] — the GitHub username.

```
def reformat_github_common(item, github_client):
    """Helper function which tweaks the data format of the parts of Issues and
    PRs which are common so that they better match Pagure
    """
    # Update reporter:
    # Search for the user
    reporter = github_client.get_user(item["user"]["login"])
    # Update the reporter field in the message (to match Pagure format)
    if reporter.name and reporter.name != "None":
        item["user"]["fullname"] = reporter.name
    else:
        item["user"]["fullname"] = item["user"]["login"]
```

fedmsg.d/sync2jira.py
Bala-Sakabattula: I guess fix version and tags are not part of PR as we can see the PR intermediary class doesn't have tags or fixVersion attributes

webbnh: OK.  That won't be a problem for now, but eventually someone might notice (e.g., PRs _can_ have labels and milestones, like Issues do, so it would potentially make sense for Sync2Jira to process them...).

The important thing for now is to remember that Sync2Jira is not completely consistent between Issues and PRs and to be careful of claims that it is....  🥴

fedmsg.d/sync2jira.py
Bala-Sakabattula: I guess PR webhook handling follows two paths, and _update_transition behaves correctly in both.


1. github.pull_request path — creates a PR object with status=None. Since _update_transition checks issue.status == "Closed", the condition always fails and it's a no-op. This is fine because _update_pr_transitions already handles merge_transition and link_transition.

2. github.issues path (PR detected via "pull_request" in body["issue"]) — creates an Issue object whose status reflects the actual PR state ("Open" or "Closed"). Here, _update_transition is live.

    
    - Merged PR: _update_pr_transitions fires merge_transition first (e.g., transitions Jira to "Closed"). When _update_transition runs next, the Jira issue is already in the target state, so the guard existing.fields.status.name.upper() != closed_status.upper() is False — no duplicate transition.
    
    - Closed without merge: merge_transition skips (because "merged" not in pr.suffix). But _update_transition sees status == "Closed" and fires, transitioning the Jira issue. This covers a case that merge_transition alone cannot handle.

So if someone configures {'transition': 'Closed'} alongside {'merge_transition': 'Closed'} in pr_updates, the two mechanisms complement each other without conflict.

fedmsg.d/sync2jira.py
Bala-Sakabattula: Okay than I am using label and milestone for tags and fixversion let it process it for PR as well..

tests/test_downstream_issue.py
Bala-Sakabattula: do we require explicitly issue_updates as parameter as if we don't provide parameter it will provide it by default so its covered right ..  and empty updates_key value is checked before this function is called in the code ..

fedmsg.d/sync2jira.py
webbnh: Do we have one or more tests which exercise this?  Should we?...  🥴

tests/test_downstream_issue.py
webbnh: The tests should determine two things:
- Does the CUT (in particular, the function interface) do everything that it is supposed to do?
- Does the CUT behave predictably for all (testable) scenarios?

The first one is validating "the specification" and "the interface contract".  (In our case, these are mostly conceptual things, rather than something written down, but we can nevertheless embody them in the test code.)

The second one is validated by getting the coverage to as close to 100% as practical.  If we can show that we have 100% coverage, then we can claim that the tests have fully exercised the CUT and that it responds properly in every possible situation.

However, note that, while these two goals _intersect_, meeting one of them does not necessarily meet the other.  In particular, 100% code coverage, doesn't guarantee that we have fully tested the interface contract.

For instance, in our case here where we have a default parameter value, from a coverage perspective, we know that there is no difference between explicitly providing the default value for that parameter and omitting the value.  But, from an interface contract perspective, those are very different tests (even though having both of them does not increase the coverage).

Furthermore, we want to have both tests, because if, in the future, the default value were to change, then we would be left with a gap if we didn't test the previous default value.  So, from a testing resilience perspective, we should test with the default value, with the value omitted, and with at least one other value.

fedmsg.d/sync2jira.py
webbnh: So, should we have
```python
                        {'fixVersion': {'overwrite': False}},  # Sync milestone as fixVersion
                        {'transition': 'Closed'},  # Transition Jira when upstream issue closes
```
here for PRs, then?

webbnh (dismissed):
It looks good.  There are a couple of possible additions to the example config, and there are a couple of opportunities for additional testing, but if you're not interested, I'd say it's good to go.

fedmsg.d/sync2jira.py
Bala-Sakabattula: We can have let me add them ...

Bala-Sakabattula:
@webbnh I added test cases for transition flow with three cases scenarios as well for maybe_convert_markdown

webbnh (dismissed):
It looks good, but, unfortunately, you've got merge conflicts to resolve.  😬

webbnh (approved):
👍

PR #477, opened 2026-05-29
PR-to-issue fixes for story points and priority

Bala-Sakabattula:
Problem:
PRs are syncing as issues, but the story points and priority fields are not being synced.

Fix:
In the upstream layer, we already fetch GitHub project values such as `estimate` (used as story points) and `priority` for issues. The same logic has now been extended for PRs as well, ensuring that both story points and priority are synced correctly.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -1295,7 +1295,8 @@ def _update_github_project_fields(
     for name, values in github_project_fields.items():
         if name not in dir(issue):
             log.error(
-                f"Configuration error: github_project_field key, {name!r}, is not in issue object."
+                "Configuration error: github_project_field key, %r, is not in issue object.",
+                name,
             )
             continue
 
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -1295,7 +1295,7 @@ def _update_github_project_fields(
     for name, values in github_project_fields.items():
         if name not in dir(issue):
             log.error(
-                f"Configuration error: github_project_field key, {name:r}, is not in issue object."
+                f"Configuration error: github_project_field key, {name!r}, is not in issue object."
             )
             continue
 
--- a/sync2jira/upstream_pr.py
+++ b/sync2jira/upstream_pr.py
@@ -19,7 +19,7 @@
 
 import logging
 
-from github import Github, UnknownObjectException
+from github import UnknownObjectException
 
 import sync2jira.intermediary as i
 import sync2jira.upstream_issue as u_issue
@@ -44,9 +44,7 @@ def handle_github_message(body, config, suffix):
     pr = body["pull_request"]
     if not u_issue.passes_github_filters(pr, config, upstream, item_type="PR"):
         return None
-    token = config["sync2jira"].get("github_token")
-    headers = {"Authorization": "token " + token} if token else {}
-    github_client = Github(token, retry=5)
+    headers, github_client = u_issue._github_client(config)
     reformat_github_pr(pr, upstream, github_client)
     u_issue.add_project_values(pr, upstream, headers, config, "pr_updates")
     return i.PR.from_github(upstream, pr, suffix, config, body.get("action"))
@@ -61,9 +59,7 @@ def github_prs(upstream, config):
     :returns: a generator for GitHub PR objects
     :rtype: Generator[sync2jira.intermediary.PR]
     """
-    github_client = Github(config["sync2jira"]["github_token"])
-    token = config["sync2jira"].get("github_token")
-    headers = {"Authorization": "token " + token} if token else {}
+    headers, github_client = u_issue._github_client(config)
     for pr in u_issue.generate_github_items("pulls", upstream, config):
         reformat_github_pr(pr, upstream, github_client)
         u_issue.add_project_values(pr, upstream, headers, config, "pr_updates")
--- a/sync2jira/upstream_pr.py
+++ b/sync2jira/upstream_pr.py
@@ -44,7 +44,7 @@ def handle_github_message(body, config, suffix):
     pr = body["pull_request"]
     if not u_issue.passes_github_filters(pr, config, upstream, item_type="PR"):
         return None
-    headers, github_client = u_issue._github_client(config)
+    headers, github_client = u_issue.get_github_client(config)
     reformat_github_pr(pr, upstream, github_client)
     u_issue.add_project_values(pr, upstream, headers, config, "pr_updates")
     return i.PR.from_github(upstream, pr, suffix, config, body.get("action"))
@@ -59,7 +59,7 @@ def github_prs(upstream, config):
     :returns: a generator for GitHub PR objects
     :rtype: Generator[sync2jira.intermediary.PR]
     """
-    headers, github_client = u_issue._github_client(config)
+    headers, github_client = u_issue.get_github_client(config)
     for pr in u_issue.generate_github_items("pulls", upstream, config):
         reformat_github_pr(pr, upstream, github_client)
         u_issue.add_project_values(pr, upstream, headers, config, "pr_updates")
--- a/sync2jira/upstream_pr.py
+++ b/sync2jira/upstream_pr.py
@@ -45,8 +45,10 @@ def handle_github_message(body, config, suffix):
     if not u_issue.passes_github_filters(pr, config, upstream, item_type="PR"):
         return None
     token = config["sync2jira"].get("github_token")
+    headers = {"Authorization": "token " + token} if token else {}
     github_client = Github(token, retry=5)
     reformat_github_pr(pr, upstream, github_client)
+    u_issue.add_project_values(pr, upstream, headers, config, "pr_updates")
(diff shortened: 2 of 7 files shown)


sync2jira/downstream_issue.py
```diff
@@ -1295,7 +1295,7 @@ def _update_github_project_fields(
     for name, values in github_project_fields.items():
         if name not in dir(issue):
             log.error(
-                f"Configuration error: github_project_field key, {name:r}, is not in issue object."
+                f"Configuration error: github_project_field key, {name!r}, is not in issue object."
```
webbnh: This is an acceptable fix, but replacing the f-string with a string using %-format specifiers (`%r`) to allow the formatting to be deferred would be better practice, e.g.:
```python
            log.error(
                "Configuration error: github_project_field key, %r, is not in issue object.",
                name
            )
```

And, the same should be applied to the other `log` calls in the function (and/or module).

sync2jira/upstream_pr.py
```diff
@@ -60,8 +62,11 @@ def github_prs(upstream, config):
     :rtype: Generator[sync2jira.intermediary.PR]
     """
     github_client = Github(config["sync2jira"]["github_token"])
+    token = config["sync2jira"].get("github_token")
+    headers = {"Authorization": "token " + token} if token else {}
     for pr in u_issue.generate_github_items("pulls", upstream, config):
         reformat_github_pr(pr, upstream, github_client)
+        u_issue.add_project_values(pr, upstream, headers, config, "pr_updates")
         yield i.PR.from_github(upstream, pr, "open", config)
```
webbnh: This (mostly) same code appears in two places (i.e., [here](https://github.com/Bala-Sakabattula/Sync2Jira/blob/940b014fe736ddc76a3b1cde39f11b51f78dedea/sync2jira/upstream_issue.py#L218-L234)), and nearly-identical code appears in two other places ([here](https://github.com/Bala-Sakabattula/Sync2Jira/blob/940b014fe736ddc76a3b1cde39f11b51f78dedea/sync2jira/upstream_issue.py#L201-L206) and [here](https://github.com/Bala-Sakabattula/Sync2Jira/blob/940b014fe736ddc76a3b1cde39f11b51f78dedea/sync2jira/upstream_pr.py#L64-L70)).

I'm not sure it's practical to refactor them into common code, but, failing that, it would be best if all four instances were all as close to identical as possible (i.e., keep the same statements in the same order; use the same idioms in each statement if appropriate, specify the same arguments, etc.) so that we can quickly compare them to make sure that nothing is missing.

tests/test_upstream_issue.py
```diff
@@ -1009,6 +1009,164 @@ def test_add_project_values_storypoints(self, mock_requests_post):
...
+                [
+                    {"fieldName": {"name": "Priority"}, "name": "Critical"},
+                    {"fieldName": {"name": "Size"}, "name": "Medium"},
+                ],
+                3,
+                "Critical",
+            ),
+            (
+                "Priority only, no storypoints config",
+                {
```
webbnh, line 1099: You have a priority-only scenario...do we want a storypoints-only scenario?

tests/test_upstream_issue.py
```diff
@@ -1009,6 +1009,164 @@ def test_add_project_values_storypoints(self, mock_requests_post):
...
+                    "priority": {"gh_field": "Priority"},
+                    "storypoints": {
+                        "gh_field": "Size",
+                        "options": {"Small": 1, "Medium": 3, "Large": 8},
+                    },
+                },
+                [
+                    {"fieldName": {"name": "Priority"}, "name": "Critical"},
+                    {"fieldName": {"name": "Size"}, "name": "Medium"},
```
webbnh: This is the same order as lines 1075-1076...just for variety, consider reversing the order here.

tests/test_upstream_pr.py
```diff
@@ -207,6 +207,198 @@ def test_github_issues(
         self.mock_github_pr.get_issue_comments.assert_any_call()
         self.assertEqual(response[0], "Successful Call!")
 
+    @mock.patch("sync2jira.intermediary.PR.from_github")
+    @mock.patch("sync2jira.upstream_issue.requests.post")
+    @mock.patch(PATH + "Github")
+    @mock.patch(PATH + "u_issue.get_all_github_data")
+    def test_github_prs_with_storypoints(
```
webbnh: Rather than mocking `requests.post()` and `get_all_github_data()`, which aren't called directly by the CUT, consider mocking `add_project_values()` and `generate_github_items()`, instead.  (I'm assuming that `add_project_values()` is adequately tested elsewhere and doesn't need to be tested here; and, unless we have no other testing for it, I would mock-out `reformat_github_pr()`, too.)

The CUT is really very simple (it's basically just a single loop...it's almost not worth testing...).  Basically, you have three scenarios -- `generate_github_items()` returns zero items, one item, and more than one item -- and, as a result, `github_prs()` should produce zero, one, and more items, respectively.

I don't think we need to test the side-effects of the subroutine calls in this test -- those should be tested elsewhere.  I think that all we really care about is whether we get the right number of intermediate objects back based on what `generate_github_items()` produces.

tests/test_upstream_pr.py
```diff
@@ -207,6 +207,198 @@ def test_github_issues(
...
+        self.mock_github_client.get_repo.assert_called_with("org/repo")
+        self.mock_github_repo.get_pull.assert_called_with(number="1234")
+        self.mock_github_pr.get_issue_comments.assert_any_call()
+        self.assertEqual(response[0], "Successful Call!")
+
+    @mock.patch("sync2jira.intermediary.PR.from_github")
+    @mock.patch("sync2jira.upstream_issue.requests.post")
+    @mock.patch(PATH + "Github")
+    @mock.patch(PATH + "u_issue.get_all_github_data")
+    def test_github_prs_with_priority(
```
webbnh: If you take my suggestion for reworking `test_github_prs_with_storypoints()`, then this test can be removed (and `test_github_prs_with_storypoints()` should probably be renamed, since the "storypoints" part won't matter or be tested).

webbnh (dismissed):
Bala, this looks great.  I just have some small things that it would be nice if you would fix.

- The `logging` call that you fixed would be best with a %-style format specifier, and you might as well fix the other nearby `logging` calls while you're at it.
- There's a blob of code which appears four times in the sources...it would be nice if all four were kept very consistent with the others, since it's probably not practical to make them common.
- The `github_prs()` tests are probably testing way too much.
- And, there are a couple of nits.

sync2jira/upstream_pr.py
Bala-Sakabattula: For headers and github client part we can make it different function and call it from the all the four places.

Bala-Sakabattula:
@webbnh I made the changes as per suggestion can you please review it again..

sync2jira/upstream_pr.py
```diff
@@ -44,9 +44,7 @@ def handle_github_message(body, config, suffix):
     pr = body["pull_request"]
     if not u_issue.passes_github_filters(pr, config, upstream, item_type="PR"):
         return None
-    token = config["sync2jira"].get("github_token")
-    headers = {"Authorization": "token " + token} if token else {}
-    github_client = Github(token, retry=5)
+    headers, github_client = u_issue._github_client(config)
```
webbnh: Given that we have cross-module references, `github_client()` should _not_ be named with an underscore prefix!

tests/test_upstream_pr.py
```diff
@@ -208,199 +209,56 @@ def test_github_issues(
...
+                ]
+
+                result = list(
+                    u.github_prs(upstream="org/repo", config=self.mock_config)
+                )
+
+                self.assertEqual(len(result), expected_count, description)
+                self.assertEqual(mock_reformat.call_count, expected_count)
+                self.assertEqual(mock_add_project.call_count, expected_count)
+                self.assertEqual(mock_pr_from_github.call_count, expected_count)
```
webbnh: I'm inclined to think that these are implementation details which we don't really want to require -- whether these functions are called directly or a specific number of times is not part of the interface contract.

If you really want to exercise this, the assertions should be checking _what_ was returned, not _how_ it was instantiated.

Otherwise, we're locking ourselves into a particular implementation, which is counterproductive.

sync2jira/upstream_pr.py
webbnh: That worked out very nicely.  Thanks!

webbnh (dismissed):
The refactor turned out very nicely, thanks!

I've flagged a couple of things which I think you should consider, but I'm approving anyway.

sync2jira/upstream_pr.py
Bala-Sakabattula: changed the name to get_github_client ad if we remove underscore there will be conflict with local variable.

tests/test_upstream_pr.py
Bala-Sakabattula: replaced the three call-count assertions with a single assertion on the returned values — verifying that github_prs() yields exactly the PR objects produced by PR.from_github.

webbnh (approved):
👍

PR #480, opened 2026-06-02
Fix assignee being cleared when upstream user has no fullname

Bala-Sakabattula:
**Summary**

- Fix a bug in _update_assignee() where an upstream GitHub assignee without a fullname in their profile would cause the downstream Jira assignee to be incorrectly cleared. The root cause was that us_exists checked for fullname (optional) instead of login (always present), so a valid assignee with no display name was treated as "no assignee", triggering a de-assignment.
- When login exists but fullname is absent, assign_user() is now called to resolve the user via LDAP/email lookup rather than clearing the Jira assignee.
- Expanded test_update_assignee_all with a new upstream variant (login present, no fullname key) and updated existing test data to include login fields, reflecting the corrected us_exists semantics.

### DIFF
--- a/sync2jira/downstream_issue.py
+++ b/sync2jira/downstream_issue.py
@@ -1230,7 +1230,7 @@ def _update_assignee(client, existing, issue, overwrite):
     """
 
     us_exists = bool(
-        issue.assignee and issue.assignee[0] and issue.assignee[0].get("fullname")
+        issue.assignee and issue.assignee[0] and issue.assignee[0].get("login")
     )
     assignee = existing.fields.assignee
     ds_exists = bool(assignee) and _jira_user_display_label(assignee) is not None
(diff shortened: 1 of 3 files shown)


sync2jira/downstream_issue.py
```diff
@@ -1239,14 +1239,19 @@ def _update_assignee(client, existing, issue, overwrite):
...
-            # the upstream one.
-            un = issue.assignee[0]["fullname"]
-            dn = _jira_user_display_label(assignee)
-            update = un != dn and remove_diacritics(un) != dn
+            un = issue.assignee[0].get("fullname")
+            if un:
+                # Overwrite the downstream assignment only if it is different
+                # from the upstream one.
+                dn = _jira_user_display_label(assignee)
+                update = un != dn and remove_diacritics(un) != dn
```
webbnh: Now that we're on Jira Cloud, I suspect that we should remove the diacritics from the downstream name as well for the fallback comparision:
```suggestion
                dn = _jira_user_display_label(assignee)
                update = un != dn and remove_diacritics(un) != remove_diacritics(dn)
```

sync2jira/downstream_issue.py
```diff
@@ -1230,7 +1230,7 @@ def _update_assignee(client, existing, issue, overwrite):
     """
 
     us_exists = bool(
-        issue.assignee and issue.assignee[0] and issue.assignee[0].get("fullname")
+        issue.assignee and issue.assignee[0] and issue.assignee[0].get("login")
     )
```
webbnh, line 1234: It seems unlikely that `issue.assignee[0]` would be true-y when `issue.assignee[0].get("login")`was not...but, I guess, from a belt-and-suspenders perspective, the extra check is OK.

tests/test_downstream_issue.py
```diff
@@ -1714,10 +1718,12 @@ def test_update_assignee_all(self, mock_client, mock_assign_user):
                 (33, False),
                 #       - upstream assignee has a fullname of None: called with remove_all=False
                 (34, False),
```
webbnh: Given that you updated the comments for scenarios 14 and 24, you should probably do the same for 34, 44, 54, and 64.

webbnh (dismissed):
Looks good.  I've flagged a nit, offered a suggestion, and made another comment.

sync2jira/downstream_issue.py
Bala-Sakabattula: okay should I remove the extra check .`get("login")`??

tests/test_downstream_issue.py
Bala-Sakabattula: Changed the comments to the following comment
` upstream assignee has login but fullname is None: not called`

Bala-Sakabattula:
@webbnh I made the changes as per suggestion.. can you kindly please review again??

sync2jira/downstream_issue.py
webbnh: At this point, nah...let's just merge this -- you can remove it some other time.

webbnh (approved):
👍

PR #491, opened 2026-07-23
handling PR updates

Bala-Sakabattula:
Handling the PR comment updates coming through github.issue_comments topic. Initially it checked in upstream_issue handler that if the its PR and if its not closed than it would ignore the PR events now removed the check so it will sync the comments from PR coming through issue handler and also in checking the github passes filter passing correct item type..

### DIFF
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -205,11 +205,13 @@ def handle_github_message(body, config, is_pr=False):
     upstream = "{owner}/{repo}".format(owner=owner, repo=repo)
 
     issue = body["issue"]
-    if not passes_github_filters(issue, config, upstream, item_type="issue"):
-        return None
-    if is_pr and not issue.get("closed_at"):
+    item_type = "PR" if is_pr else "issue"
+    if not passes_github_filters(issue, config, upstream, item_type=item_type):
         log.debug(
-            "%r is a pull request.  Ignoring.", issue.get("html_url", "<missing URL>")
+            "%s %s %s does not pass filters",
+            item_type,
+            upstream,
+            issue.get("html_url", "<missing URL>"),
         )
         return None
 
--- a/sync2jira/upstream_issue.py
+++ b/sync2jira/upstream_issue.py
@@ -208,10 +208,10 @@ def handle_github_message(body, config, is_pr=False):
     item_type = "PR" if is_pr else "issue"
     if not passes_github_filters(issue, config, upstream, item_type=item_type):
         log.debug(
-            "%s %s %s does not pass filters",
+            "%s %s#%s does not pass filters",
             item_type,
             upstream,
-            issue.get("html_url", "<missing URL>"),
+            issue.get("number", "?"),
         )
         return None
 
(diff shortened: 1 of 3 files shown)


sync2jira/upstream_issue.py
```diff
@@ -205,11 +205,13 @@ def handle_github_message(body, config, is_pr=False):
...
-    if is_pr and not issue.get("closed_at"):
+    item_type = "PR" if is_pr else "issue"
+    if not passes_github_filters(issue, config, upstream, item_type=item_type):
         log.debug(
-            "%r is a pull request.  Ignoring.", issue.get("html_url", "<missing URL>")
+            "%s %s %s does not pass filters",
+            item_type,
+            upstream,
+            issue.get("html_url", "<missing URL>"),
         )
```
webbnh, line 215: Logging `upstream` seems redundant, since that information should be in the `"html_url"` value.

But, rather than logging the URL, consider something similar to the call at line 239.

tests/test_upstream_issue.py
```diff
@@ -502,21 +502,23 @@ def test_handle_github_message_not_in_mapped(
 
     @mock.patch(PATH + "Github")
     @mock.patch("sync2jira.intermediary.Issue.from_github")
-    def test_handle_github_message_pull_request(
+    def test_handle_github_message_pull_request_not_in_sync(
```
webbnh, line 505: These changes are a good improvement, but they raise the question of whether we shouldn't have more tests for the filtering.  (Should we have testing specifically for `passes_github_filters()`?)

webbnh (dismissed):
The changes look good, but they seem a little light on testing.  (And, I flagged one other nit.)

sync2jira/upstream_issue.py
Bala-Sakabattula: adding the issue number in the logs

tests/test_upstream_issue.py
Bala-Sakabattula: https://github.com/release-engineering/Sync2Jira/blob/main/tests/test_upstream_issue.py#L1181-1268
Test case exists for passes_github_filters() function ....want  me to add more test cases ??

Bala-Sakabattula:
@webbnh I added test cases for passes_github_filters function and also instead of html url logging issue number

tests/test_upstream_issue.py
```diff
@@ -1180,6 +1180,58 @@ def test_add_project_values_pr(self, mock_requests_post):
...
+        good_item = {
+            "labels": [{"name": "custom_tag"}],
+            "milestone": {"number": 1},
+            "filter1": "filter1",
+        }
+
+        # upstream not in mapped_repos → None
+        self.assertIsNone(
+            u.passes_github_filters(
+                good_item, self.mock_config, "other/repo", item_type="issue"
```
webbnh, line 1200: If you make an edit for some other reason, you should use `upstream` here instead of the `"other/repo"` literal (e.g., see line 1210).

webbnh (approved):
Looks good.  (Although, I did find another nit, but we can ignore it....)

> they seem a little light on testing.

That comment was inaccurate:  the PR did include reasonable testing for the lines that you actually changed.

But, I think you agree that `passes_github_filters()` wasn't adequately tested (i.e., because we're now sending PRs through it, when we weren't before), so thank you for the additions!

tests/test_upstream_issue.py
Bala-Sakabattula: ohhhh okay!! I will change it..

PR #496, opened 2026-08-05
comment limit fix

Bala-Sakabattula:
The Jira Python library returns only the first 100 comments by default. Therefore, if an issue has more than 100 comments, we need to fetch all comments using pagination before performing any checks.

Reference: https://github.com/pycontribs/jira/issues/1938

### DIFF
--- a/sync2jira/downstream_pr.py
+++ b/sync2jira/downstream_pr.py
@@ -32,6 +32,7 @@ from sync2jira.intermediary import Issue, matcher
 log = logging.getLogger("sync2jira")
 
 UPDATES_KEY = "pr_updates"
+PAGE_SIZE = 100
 
 
 def format_comment(pr, pr_suffix, client):
@@ -92,16 +93,15 @@ def comment_exists(client, existing: JIRAIssue, new_comment):
     :param String new_comment: Formatted comment we're looking for
     :returns: Nothing
     """
-    page_size = 100
     start = 0
     while True:
-        batch = client.comments(existing, start_at=start, max_results=page_size)
+        batch = client.comments(existing, start_at=start, max_results=PAGE_SIZE)
         for comment in batch:
             if new_comment == comment.body:
                 return True
-        if len(batch) < page_size:
+        if len(batch) < PAGE_SIZE:
             break  # last page; comment not found
-        start += page_size
+        start += PAGE_SIZE
     return False
 
 
--- a/sync2jira/downstream_pr.py
+++ b/sync2jira/downstream_pr.py
@@ -27,12 +27,12 @@ from jira.client import ResultList
 
 # Local Modules
 import sync2jira.downstream_issue as d_issue
+from sync2jira.downstream_issue import JIRA_COMMENTS_PAGE_SIZE
 from sync2jira.intermediary import Issue, matcher
 
 log = logging.getLogger("sync2jira")
 
 UPDATES_KEY = "pr_updates"
-PAGE_SIZE = 100
 
 
 def format_comment(pr, pr_suffix, client):
@@ -95,13 +95,15 @@ def comment_exists(client, existing: JIRAIssue, new_comment):
     """
     start = 0
     while True:
-        batch = client.comments(existing, start_at=start, max_results=PAGE_SIZE)
+        batch = client.comments(
+            existing, start_at=start, max_results=JIRA_COMMENTS_PAGE_SIZE
+        )
         for comment in batch:
             if new_comment == comment.body:
                 return True
-        if len(batch) < PAGE_SIZE:
+        if len(batch) < JIRA_COMMENTS_PAGE_SIZE:
             break  # last page; comment not found
-        start += PAGE_SIZE
+        start += JIRA_COMMENTS_PAGE_SIZE
     return False
 
 
--- a/sync2jira/downstream_pr.py
+++ b/sync2jira/downstream_pr.py
@@ -92,11 +92,16 @@ def comment_exists(client, existing: JIRAIssue, new_comment):
     :param String new_comment: Formatted comment we're looking for
     :returns: Nothing
     """
-    # Grab and loop over comments
-    comments = client.comments(existing)
-    for comment in comments:
-        if new_comment == comment.body:
-            return True
+    page_size = 100
+    start = 0
+    while True:
+        batch = client.comments(existing, start_at=start, max_results=page_size)
+        for comment in batch:
+            if new_comment == comment.body:
+                return True
+        if len(batch) < page_size:
+            break  # last page; comment not found
+        start += page_size
     return False
 
 
--- a/tests/test_downstream_issue.py
+++ b/tests/test_downstream_issue.py
@@ -1569,16 +1569,21 @@ class TestDownstreamIssue(unittest.TestCase):
                 else:
                     mock_client.add_comment.assert_not_called()
 
+    @mock.patch(PATH + "_truncate_jira_text")
     @mock.patch(PATH + "_comment_format")
     @mock.patch(PATH + "_comment_matching")
     @mock.patch("jira.client.JIRA")
     def test_update_comments(
-        self, mock_client, mock_comment_matching, mock_comment_format
+        self,
+        mock_client,
+        mock_comment_matching,
+        mock_comment_format,
+        mock_truncate_jira_text,
     ):
         """
-        Single-page case: Jira returns fewer than 100 comments, so the
-        paginator stops after one request.  The one upstream comment that
-        is not yet in Jira is formatted and added.
+        Single-page case: Jira returns fewer than 100 comments so the
+        paginator stops after one request.  The upstream comment that is
+        not yet in Jira is formatted and added to the Jira issue.
         """
         upstream_comment = {"id": "1", "body": "hello", "changed": None}
         self.mock_issue.comments = [upstream_comment]
@@ -1588,6 +1593,7 @@ class TestDownstreamIssue(unittest.TestCase):
             upstream_comment
         ]  # not matched → stays pending
         mock_comment_format.return_value = "formatted_body"
+        mock_truncate_jira_text.return_value = "truncated_body"
 
         d._update_comments(
             client=mock_client, existing=self.mock_downstream, issue=self.mock_issue
@@ -1596,11 +1602,10 @@ class TestDownstreamIssue(unittest.TestCase):
         mock_client.comments.assert_called_once_with(
             self.mock_downstream, start_at=0, max_results=100
         )
-        mock_comment_matching.assert_called_once_with(
-            [upstream_comment], ["jira_comment"], self.mock_issue.url
+        # Verify the right content reached the Jira client
+        mock_client.add_comment.assert_called_once_with(
+            self.mock_downstream, "truncated_body"
         )
-        mock_comment_format.assert_called_once_with(upstream_comment)
-        mock_client.add_comment.assert_called_once_with(self.mock_downstream, mock.ANY)
 
     @mock.patch(PATH + "_comment_format")
     @mock.patch(PATH + "_comment_matching")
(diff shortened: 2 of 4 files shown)


sync2jira/downstream_pr.py
```diff
@@ -92,11 +92,16 @@ def comment_exists(client, existing: JIRAIssue, new_comment):
     :param String new_comment: Formatted comment we're looking for
     :returns: Nothing
     """
-    # Grab and loop over comments
-    comments = client.comments(existing)
-    for comment in comments:
-        if new_comment == comment.body:
-            return True
+    page_size = 100
```
webbnh: You've defined `page_size` three times in three separate places; consider using a common definition, since I think it's a constant with the same concept/usage in all three cases.

tests/test_downstream_issue.py
```diff
@@ -1576,28 +1576,120 @@ def test_update_comments(
...
+        is not yet in Jira is formatted and added.
         """
-        # Set up return values
-        mock_client.comments.return_value = "mock_comments"
-        mock_comment_matching.return_value = ["mock_comments_d"]
-        mock_comment_format.return_value = "mock_comment_body"
+        upstream_comment = {"id": "1", "body": "hello", "changed": None}
+        self.mock_issue.comments = [upstream_comment]
+
+        mock_client.comments.return_value = ["jira_comment"]  # 1 item < 100 → last page
```
webbnh, line 1591: I wonder if we should have a _zero items returned_ case.  (E.g., if Jira had exactly a hundred items, then after the first batch we would request a second batch which would come back empty.)

tests/test_downstream_issue.py
```diff
@@ -1576,28 +1576,120 @@ def test_update_comments(
...
+        mock_client.comments.assert_called_once_with(
+            self.mock_downstream, start_at=0, max_results=100
         )
-        mock_comment_format.assert_called_with("mock_comments_d")
-        mock_client.add_comment.assert_called_with(
-            self.mock_downstream, "mock_comment_body"
+        mock_comment_matching.assert_called_once_with(
+            [upstream_comment], ["jira_comment"], self.mock_issue.url
+        )
+        mock_comment_format.assert_called_once_with(upstream_comment)
```
webbnh: I understand why you're including these assertions, and I agree with the sentiment, but I wonder if this isn't making the test too tightly-coupled to the existing implementation, such that it will restrict future enhancements or refactoring.

It's probably OK to use the `mock_client` to get insight into the CUT behavior, because this is an external dependency, but, instead of checking the calls and arguments to `mock_comment_matching` and `mock_comment_format`, consider looking at the arguments to `mock_client.add_comment()` to see if they match expectations (e.g., that the comment has `mock_comment_format.return_value`) -- this checks the "what" without constraining the "how" too much.  (That is, the point of this function is to call the Jira client the right number of times with the "right stuff"; we don't care too much about calls to internal functions like `_comment_matching()` and `comment_format()` because those are implementation details leading up to the external call.)

The same consideration should be applied to the other tests, as well.

tests/test_downstream_issue.py
```diff
@@ -1576,28 +1576,120 @@ def test_update_comments(
...
-        mock_comment_matching.return_value = ["mock_comments_d"]
-        mock_comment_format.return_value = "mock_comment_body"
+        upstream_comment = {"id": "1", "body": "hello", "changed": None}
+        self.mock_issue.comments = [upstream_comment]
+
+        mock_client.comments.return_value = ["jira_comment"]  # 1 item < 100 → last page
+        mock_comment_matching.return_value = [
+            upstream_comment
+        ]  # not matched → stays pending
+        mock_comment_format.return_value = "formatted_body"
```
webbnh, line 1595: I suspect that we should be mocking `_truncate_jira_text()` in addition to `_comment_format()`....

tests/test_downstream_issue.py
```diff
@@ -1576,28 +1576,120 @@ def test_update_comments(
...
+            [upstream_comment], ["jira_comment"], self.mock_issue.url
+        )
+        mock_comment_format.assert_called_once_with(upstream_comment)
+        mock_client.add_comment.assert_called_once_with(self.mock_downstream, mock.ANY)
+
+    @mock.patch(PATH + "_comment_format")
+    @mock.patch(PATH + "_comment_matching")
+    @mock.patch("jira.client.JIRA")
+    def test_update_comments_no_upstream_comments(
+        self, mock_client, mock_comment_matching, mock_comment_format
```
webbnh: We don't reference the `_comment_format()` mock in the test...should it be omitted?  (If not, should we also be mocking `_truncate_jira_text()`?)

Ditto for `test_update_comments_early_stop()` and `test_update_comments_multi_page()`

tests/test_downstream_issue.py
```diff
@@ -1576,28 +1576,120 @@ def test_update_comments(
...
+
+    @mock.patch(PATH + "_comment_format")
+    @mock.patch(PATH + "_comment_matching")
+    @mock.patch("jira.client.JIRA")
+    def test_update_comments_multi_page(
+        self, mock_client, mock_comment_matching, mock_comment_format
+    ):
+        """
+        Multi-page case: upstream comment is not found on the first two
+        full pages but is found on the third (short) page.
```
webbnh, line 1651: Do we have (and do we want) a multi-page case where the comment is not found (and is therefore added)?

tests/test_downstream_pr.py
```diff
@@ -285,39 +285,69 @@ def test_update_jira_issue_exists(
...
+        self.mock_client.comments.assert_called_once_with(
+            "mock_existing", start_at=0, max_results=100
+        )
+        self.assertFalse(response)
 
     def test_comment_exists_true(self):
         """
-        This function tests 'comment_exists' where the comment exists
+        Single-page case: comment found → returns True immediately.
         """
```
webbnh, line 308: I'm not sure that this test case is testing very much -- in particular, I don't see how it is proving "immediately".

Perhaps if the `comments()` mock returned 100 items, then the assertion that it was called only once would be a little more meaningful.  But, with it returning the "last" page, there is no reason for the CUT to call it a second time, regardless of whether it finds the comment or not.

webbnh (dismissed):
This looks good, @Bala-Sakabattula.  However, I found a bunch of small items that you might want to consider addressing.

tests/test_downstream_issue.py
Bala-Sakabattula: We added this case if the comment is not found in the first two batch of comments its querying multi batches ...

tests/test_downstream_issue.py
Bala-Sakabattula: I was under the impression that, based on the cases `test_update_comments_no_upstream`, `test_update_comments_early_stop()`, and `test_update_comments_multi_page()`, the main goal is to verify whether Jira is queried or not. We don’t necessarily need to test formatting, truncation, or adding parts of the response.

Should we include those additional checks as well?

tests/test_downstream_issue.py
webbnh: But, control will (hopefully...) follow a _different_ path if the comment is not found at all.  It's true that the path it would follow overlaps with the paths followed if the comment is found on a later page and it overlaps the path followed if there is only one page which doesn't contain the comment, but from a testing perspective these are distinct cases:  we want the code to behave properly regardless of whether the comment is found or not and regardless of the number of pages searched...which means that there are at least four cases to test (and, arguably, there are six or eight):  the comment is or is not found on the only page or on the last-of-several pages.

Once again, adding this test case will not increase the code coverage, but it would increase the _test_ coverage.  This is because the implementation is clever and efficient...but the tests are there to define and regulate the overall _behavior_ and not the details of the implementation.

tests/test_downstream_issue.py
webbnh: > the main goal is to verify whether Jira is queried or not.

Actually, the goal is to ensure that the CUT conforms to its design specification, including externally observable behaviors.  One of those behaviors would be the calls it makes to Jira.

> We don’t necessarily need to test formatting, truncation, or adding parts of the response.

In terms of testing _this_ function, your statement is correct (although, those things should also be tested, but separately), and that is **why** they should be consistently mocked out, here.  (The code as it stands is mocking one and driving the CUT to actually call the other.)

We can redefine the "unit" to _include_ those subroutines, in which case we should not mock either one; or, we can limit the "unit" to just the body of `_update_comments()`.  Including one of the subroutines in the unit and excluding the other seems weird.  Moreover, mocking out a subroutine without interacting with the mock in any way (to set return values, to observe the arguments, et al.) is also weird.

Mocking the subroutines is undesirable in that it creates a tighter coupling between the test and the CUT (i.e., it is "white-box" testing, because it requires the test to know what subroutines the implementation calls).  But, if there is a gain in speed/efficiency for the test, or if it allows the test better control over the CUT, or if it enables some sort of really important observability, then that coupling is justified.

I just called this out because of the inconsistency and because we aren't actually _using_ the mock that we're creating.

Bala-Sakabattula: to maintain the consistency I removed the _comment_format and _turncate_jira_text from test_update_comments..

tests/test_downstream_pr.py
Bala-Sakabattula: changing the definition of finding the comment in first batch so it returns true..

Bala-Sakabattula:
@webbnh according to the review added missed cases and modified the existing ones as well can you please review again???

sync2jira/downstream_pr.py
```diff
@@ -32,6 +32,7 @@
 log = logging.getLogger("sync2jira")
 
 UPDATES_KEY = "pr_updates"
+PAGE_SIZE = 100
```
webbnh: Having this definition at the module level is better, but consider using instead the one from `sync2jira.downstream_issue` -- I'm thinking that we would be surprised if the two had different values, and, therefore, we should just have one, shared definition.

Also, "PAGE_SIZE" is a little bit generic, given the scope -- I recommend "JIRA_PAGE_SIZE" or even "JIRA_COMMENTS_PAGE_SIZE".

(The rule of thumb is that the length of a variable's (or symbolic constant's) name should be proportional to the size of its scope...so, for a nested block, you can get away with a single-letter, while for a package-wide global it should be very long -- this has two effects:  it ensures that the name is properly self-documenting, and it makes people tend to avoid using globals because they have such awkward and ugly names!  😁)

tests/test_downstream_issue.py
webbnh: I'm not sure what your comment here is referring to.  Looking at `test_update_comments()`, you _added_ a mock for `_truncate_jira_text()`, and the mock for `_comment_format()` is still there.  (This is fine, now -- we are mocking the CUT's dependencies, and you are using the mocks to drive the CUT's behavior.)

However, the code here (in `test_update_comments_no_upstream_comments()`) still has the inconsistent mocking -- it mocks `_comment_format()` without also mocking `_truncate_jira_text()`.  I recommend that it do the same thing that `test_update_comments()` does.

(It looks like you also corrected `test_update_comments_early_stop()`, so perhaps you just missed the function here.)

tests/test_downstream_issue.py
```diff
@@ -1690,6 +1698,98 @@ def test_update_comments_multi_page(
...
+        mock_client.add_comment.assert_called_once_with(
+            self.mock_downstream, "truncated_body"
+        )
+
+    @mock.patch(PATH + "_comment_format")
+    @mock.patch(PATH + "_comment_matching")
+    @mock.patch("jira.client.JIRA")
+    def test_update_comments_found_on_full_intermediate_page(
+        self, mock_client, mock_comment_matching, mock_comment_format
+    ):
```
webbnh: Should we omit the mock for `_comment_format()`?  The test never references it, and I don't expect the CUT to actually call it.  (Also, if we're going to include it, then we should probably also mock `_truncate_jira_text()`....)

tests/test_downstream_pr.py
```diff
@@ -304,22 +304,50 @@ def test_comment_exists_false(self):
...
         mock_comment = MagicMock()
         mock_comment.body = "mock_new_comment"
-        self.mock_client.comments.return_value = [
-            mock_comment
-        ]  # 1 item < 100 → last page
+        other_comment = MagicMock()
+        other_comment.body = "something else"
+        self.mock_client.comments.return_value = [mock_comment] + [
+            other_comment
+        ] * 99  # 1 item passes other items
```
webbnh: This is cool, but are you sure that you want the found comment to be the very first one looked at?  It might be a better test if it were the last or in the middle.

webbnh (dismissed):
Looks good -- nice additions.

I just have a few follow-ups.

tests/test_downstream_issue.py
Bala-Sakabattula: Removed it...as CUT doesn't call it..

sync2jira/downstream_pr.py
Bala-Sakabattula: I will change the name but the fetching it from downstream issue

webbnh (approved):
👍

PR #501, opened 2026-09-02
Run repository sync in background thread to prevent gateway timeouts and add sync status tracking

Bala-Sakabattula:
**Problem**

Currently, repository synchronization is performed synchronously in the main request-handling flow. Since syncing repositories can take a significant amount of time, the request does not return a response before the gateway timeout occurs.

This results in a gateway timeout even though the synchronization may still be processing.

**Changes**
Background repository synchronization

- Moved the repository synchronization work to a background thread.
- The main request-handling thread is freed immediately after starting the background job, allowing a valid response to be returned before the gateway timeout.
- This does not introduce parallel/multi-threaded repository synchronization within a single job. The background thread processes the selected repositories as part of the same sync job.

Repository-level locking

- Added repository-level locking to prevent the same repository from being synchronized by multiple user requests at the same time.
- Different repositories are still allowed to be synchronized independently.
- Once a repository sync completes or fails, the repository is removed from the active sync lock so it can be synchronized again.

Job tracking and status

- Added a job ID for each sync request.
- Added an in-memory job dictionary to maintain the status of each individual request.
- Each job tracks its current status and any error that occurs during synchronization.
- The job status is exposed through a status endpoint so the frontend can retrieve the status for the specific user's request.
- Once the sync job completes, the job information is cleaned up from the job dictionary as appropriate.

**Sync status UI**

After submitting a sync request:

- The user is redirected to an in-progress status page.
- The page polls the job status endpoint every 3 seconds.
- Repository sync status is updated based on the latest job information.
- On successful completion, the user is shown a success page.
- If the synchronization fails, the user is shown a failure page with the corresponding error message.

 **downstream.py description handling**

While updating the downstream issue description, existing.fields.description can be None.

Previously, passing a None value to difflib.unified_diff() caused the description comparison to fail.

Updated the description handling to:

- Treat a None existing description as an empty string.
- Use splitlines(keepends=True) when generating the unified diff so that line-level changes and line endings are handled correctly.

**Tests**

- Added a test case covering the downstream issue description update when the existing description is None.
- The test verifies that the description diff/update functionality works correctly with an empty existing description.

**Result**

With these changes, long-running repository synchronization no longer blocks the HTTP request until completion, preventing gateway timeout errors while providing users with real-time progress/status feedback for their sync request.

### DIFF
--- a/sync-page/event-handler.py
+++ b/sync-page/event-handler.py
@@ -1,9 +1,11 @@
 # Build-In Modules
 import logging
 import os
+import threading
+import uuid
 
 # 3rd Party Modules
-from flask import Flask, redirect, render_template, request
+from flask import Flask, jsonify, redirect, render_template, request
 
 # Local Modules
 from sync2jira.main import initialize_issues, initialize_pr, load_config
@@ -14,6 +16,12 @@ BASE_URL = os.environ["BASE_URL"]
 REDIRECT_URL = os.environ["REDIRECT_URL"]
 config = load_config()
 
+# In-memory job tracker: {job_id: {"status": str, "repos": list, "error": str|None}}
+_jobs: dict = {}
+_jobs_repo = set()
+_jobs_repo_lock = threading.Lock()
+_jobs_lock = threading.Lock()
+
 # Set up our logging
 FORMAT = "[%(asctime)s] %(levelname)s: %(message)s"
 logging.basicConfig(format=FORMAT, level=logging.INFO)
@@ -22,28 +30,76 @@ logging.basicConfig(format=FORMAT, level=logging.WARNING)
 log = logging.getLogger("sync2jira-sync-page")
 
 
+def _run_sync(job_id: str, repos: list):
+    """Run initialize_issues and initialize_pr for each repo in a background thread."""
+    try:
+        for repo_name in repos:
+            log.info(f"[job:{job_id}] Syncing repo: {repo_name}")
+            initialize_issues(config, repo_name=repo_name)
+            initialize_pr(config, repo_name=repo_name)
+            log.info(f"[job:{job_id}] Finished repo: {repo_name}")
+        with _jobs_lock:
+            _jobs[job_id]["status"] = "completed"
+    except Exception as e:
+        log.exception(f"[job:{job_id}] Sync failed")
+        with _jobs_lock:
+            _jobs[job_id]["status"] = "failed"
+            _jobs[job_id]["error"] = str(e)
+    finally:
+        # Release the repo lock only after sync completes (success or failure)
+        with _jobs_repo_lock:
+            _jobs_repo.difference_update(repos)
+
+
 @app.route("/handle-event", methods=["POST"])
 def handle_event():
     """
-    Handler for when a user wants to sync a repo
+    Handler for when a user wants to sync a repo.
+    Kicks off sync in a background thread and immediately returns an
+    in-progress page so the gateway never times out.
     """
     response = request.form
-    synced_repos = []
-    for repo_name, switch in response.items():
-        if switch == "on":
-            # Sync repo_name
-            log.info(f"Starting sync for repo: {repo_name}")
-            initialize_issues(config, repo_name=repo_name)
-            initialize_pr(config, repo_name=repo_name)
-            synced_repos.append(repo_name)
-    if synced_repos:
-        return render_template(
-            "sync-page-success.jinja",
-            synced_repos=synced_repos,
-            url=f"https://{REDIRECT_URL}",
-        )
-    else:
+    repos_to_sync = [repo for repo, switch in response.items() if switch == "on"]
+
+    if not repos_to_sync:
         return render_template("sync-page-failure.jinja", url=f"https://{REDIRECT_URL}")
+
+    with _jobs_repo_lock:
+        already_syncing = set(repos_to_sync) & _jobs_repo
+        if already_syncing:
+            return render_template(
+                "sync-page-failure.jinja",
+                url=f"https://{REDIRECT_URL}",
+                error=f"Already syncing: {', '.join(already_syncing)}",
+            )
+        _jobs_repo.update(repos_to_sync)
+
+    job_id = str(uuid.uuid4())
+    with _jobs_lock:
+        _jobs[job_id] = {"status": "in_progress", "repos": repos_to_sync, "error": None}
+
+    thread = threading.Thread(
+        target=_run_sync, args=(job_id, repos_to_sync), daemon=True
+    )
+    thread.start()
+
+    log.info(f"Started background sync job {job_id} for repos: {repos_to_sync}")
+    return render_template(
+        "sync-page-in-progress.jinja",
+        job_id=job_id,
+        synced_repos=repos_to_sync,
+        url=f"https://{REDIRECT_URL}",
+    )
+
+
+@app.route("/status/<job_id>", methods=["GET"])
+def job_status(job_id: str):
+    """Return JSON status for a background sync job."""
+    with _jobs_lock:
+        job = _jobs.get(job_id)
+    if job is None:
+        return jsonify({"status": "not_found"}), 404
+    return jsonify(job)
 
 
 @app.route("/", methods=["GET"])
--- a/sync-page/event-handler.py
+++ b/sync-page/event-handler.py
@@ -2,6 +2,7 @@
 import logging
 import os
 import threading
+import time
 import uuid
 
 # 3rd Party Modules
@@ -16,11 +17,13 @@ BASE_URL = os.environ["BASE_URL"]
 REDIRECT_URL = os.environ["REDIRECT_URL"]
 config = load_config()
 
-# In-memory job tracker: {job_id: {"status": str, "repos": list, "error": str|None}}
+# In-memory job tracker: {job_id: {"status": str, "repos": list, "error": str|None, "finished_at": float}}
 _jobs: dict = {}
 _jobs_repo = set()
 _jobs_repo_lock = threading.Lock()
 _jobs_lock = threading.Lock()
+
+JOB_TTL_SECONDS = 600  # retain terminal jobs for 10 minutes then discard
 
 # Set up our logging
 FORMAT = "[%(asctime)s] %(levelname)s: %(message)s"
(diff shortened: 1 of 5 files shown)


sync-page/event-handler.py
```diff
@@ -22,29 +30,82 @@
...
+
+@app.route("/status/<job_id>", methods=["GET"])
+def job_status(job_id: str):
+    """Return JSON status for a background sync job.
+    Cleans up the job from memory once the browser reads a terminal status.
+    """
+    with _jobs_lock:
+        job = _jobs.get(job_id)
+    if job is None:
+        return jsonify({"status": "not_found"}), 404
```
ralphbean, line 135: **[important]** Removing the job on the first terminal-status read makes that response destructive. If the response is lost, or the page is refreshed, the next poll gets `not_found` and the UI keeps polling. Jobs also remain in memory forever when nobody reads the terminal status. Could we retain terminal jobs for a bounded period and expire them independently of polling?

sync-page/templates/sync-page-in-progress.jinja
```diff
@@ -0,0 +1,136 @@
...
+              '<div class="pf-c-alert__icon"><i class="fas fa-fw fa-check-circle" aria-hidden="true"></i></div>' +
+              '<p class="pf-c-alert__title">All selected repos have been synced to JIRA.</p>' +
+              '</div>';
+            document.getElementById("action-buttons").hidden = false;
+          } else if (data.status === "failed") {
+            clearInterval(pollInterval);
+            markAllSpinners("repo-failed", "✗");
+            document.getElementById("page-title").textContent = "Sync Failed";
+            document.getElementById("overall-status").innerHTML =
+              '<div class="pf-c-alert pf-m-danger pf-m-inline" aria-label="Danger alert">' +
```
ralphbean: **[important]** `data.error` can include text from GitHub, Jira, or repository data. Inserting it through `innerHTML` lets that text become markup. Could we build the alert and assign the error through `textContent` instead?

sync-page/event-handler.py
```diff
@@ -22,29 +30,82 @@
...
+
+    log.info(f"Started background sync job {job_id} for repos: {repos_to_sync}")
+    return render_template(
+        "sync-page-in-progress.jinja",
+        job_id=job_id,
+        synced_repos=repos_to_sync,
+        url=f"https://{REDIRECT_URL}",
+    )
+
+
```
ralphbean, line 126: **[minor, non-blocking]** It would be good to add coverage for job transitions, locking, cleanup, failures, and the status endpoint. Most of the behavior introduced here currently isn’t exercised by tests.

ralphbean (changes requested):
I think we need to address the job lifecycle and error rendering before merging. I’d also like to see coverage for the new background-job behavior, but I don’t think the tests need to block this PR.

webbnh (commented):
Same thoughts as Ralph.

sync-page/event-handler.py
Bala-Sakabattula: what will be bounded period ??? I mean we need to start it since start of the sync job and after how much time we will delete it ???For now I am keeping 10mins

sync-page/event-handler.py
Bala-Sakabattula: As sync page doesn't had any test so not written any test for this cases as well...I will try to add.

sync-page/event-handler.py
Bala-Sakabattula: Used finished_at key value pair  in job dict after the 10 minutes of finishing removing the job from dict through a background thread running a every 2 mins.

ralphbean (approved):

sync-page/event-handler.py
```diff
@@ -22,29 +33,108 @@
...
+        time.sleep(120)  # check every 2 minutes
+        cutoff = time.monotonic() - JOB_TTL_SECONDS
+        with _jobs_lock:
+            if not _jobs:
+                continue
+            expired = [
+                jid
+                for jid, job in _jobs.items()
+                if job["status"] in ("completed", "failed")
+                and job["finished_at"] < cutoff
```
webbnh, line 48: I'm not sure that we need to check the status -- if the job has a "finished at" value, then it is presumably eligible for removal.

sync-page/event-handler.py
```diff
@@ -22,29 +33,108 @@
...
+            ]
+            for jid in expired:
+                _jobs.pop(jid)
+                log.debug("Expired sync job %s from memory", jid)
+
+
+threading.Thread(target=_cleanup_expired_jobs, daemon=True, name="job-cleanup").start()
```
webbnh, line 55: Rather than creating a persistent daemon thread, consider just calling `_cleanup_expired_jobs()` from `handle_event()` or some other location associated with creating a request.  We can tolerate having a small number of "dead" requests lying around in memory, and we can clean them up the next time we get a request, which means that they won't build up.  (This is simpler architecturally, and we don't risk problems from having the daemon thread lock up or running rogue.)

sync-page/event-handler.py
```diff
@@ -16,12 +17,14 @@
...
 config = load_config()
 
-# In-memory job tracker: {job_id: {"status": str, "repos": list, "error": str|None}}
+# In-memory job tracker: {job_id: {"status": str, "repos": list, "error": str|None, "finished_at": float}}
 _jobs: dict = {}
 _jobs_repo = set()
 _jobs_repo_lock = threading.Lock()
 _jobs_lock = threading.Lock()
 
+JOB_TTL_SECONDS = 600  # retain terminal jobs for 10 minutes then discard
```
webbnh, line 26: 10 minutes seems awkwardly short:  if someone wants to sync a large repo, they might start the request and then, say, go to lunch, and they won't be able to see the result when they return.  (And, what if the sync operation itself takes more than ten minutes, which seems plausible for a big repo on a day when GitHub or the network is busy.)

I don't think it will hurt us to keep things around longer -- I don't think dead requests take up much resources -- we are only concerned about reclaiming them because we don't want the "lost" storage to build up.

sync-page/event-handler.py
```diff
@@ -40,11 +65,13 @@ def _run_sync(job_id: str, repos: list):
             log.info(f"[job:{job_id}] Finished repo: {repo_name}")
         with _jobs_lock:
             _jobs[job_id]["status"] = "completed"
+            _jobs[job_id]["finished_at"] = time.monotonic()
     except Exception as e:
         log.exception(f"[job:{job_id}] Sync failed")
         with _jobs_lock:
             _jobs[job_id]["status"] = "failed"
             _jobs[job_id]["error"] = str(e)
+            _jobs[job_id]["finished_at"] = time.monotonic()
```
webbnh, line 74: I'm not necessarily objecting, but I suggest that the `"status"` field is superfluous:  the value of the `"finished_at"` value indicates whether the job is complete, and the value of the `"error"` key indicates whether there was an error or not, so we don't actually need the `"status"` value.

sync-page/event-handler.py
```diff
@@ -95,16 +127,13 @@ def handle_event():
...
     """
     with _jobs_lock:
         job = _jobs.get(job_id)
     if job is None:
         return jsonify({"status": "not_found"}), 404
-    if job["status"] in ("completed", "failed"):
-        with _jobs_lock:
-            _jobs.pop(job_id, None)
-    return jsonify(job)
+    return jsonify({k: v for k, v in job.items() if k != "finished_at"})
```
webbnh, line 136: Doing this without the `_jobs_lock` seems slightly risky.  I recommend moving this statement to line 133:  the dictionary comprehension will be conducted under the lock, and then it will be released during the `return`.

webbnh (commented):
Hi Bala, I see that this PR has been merged -- that's great.  However, if you're open to another pass at it, I have some thoughts for you to consider.

sync-page/event-handler.py
Bala-Sakabattula: We’re adding a 10-minute buffer after the sync is completed.

For a large repo, the sync may take some time. Once the sync finishes and finished_at is updated, the system will wait for another 10 minutes before deleting it.

sync-page/event-handler.py
Bala-Sakabattula: just kept extra check...

sync-page/event-handler.py
Bala-Sakabattula: We use status field in sync-page -in-progress.jinja for showing valid data in the page..

sync-page/event-handler.py
Bala-Sakabattula: we already read job detail through lock itself after fetching we are just returning by removing the finished_at field value as we are not handling time value in jinja

sync-page/event-handler.py
webbnh: Good point -- the countdown starts when the request _finishes_ not when it _starts_ -- my bad.  That relieves most of the pain.  But, since the user doesn't know how long it's going to take, they don't know when to check back to see if it's done, and they have to hit the ten-minute window.

But, as long as the completed requests do get cleaned up eventually, and as long as each one doesn't take up much space, I think we can afford to keep them around much longer than 10 minutes (like, until the next day, or for 10 or 24 hours, etc.).

sync-page/event-handler.py
webbnh: OK, that justifies keeping it.  Hopefully, having multiple sources of truth won't result in any awkward conflicts.

sync-page/event-handler.py
webbnh: OK, so this is basically making a copy of the `job` object (and omitting the `"finished_at"` key)...yeah, that seems like it should be fine.

sync-page/event-handler.py
Bala-Sakabattula: > But, since the user doesn't know how long it's going to take, they don't know when to check back to see if it's done, and they have to hit the ten-minute window.

the status page polls the status api to fetch the current state of the sync if it is in progress it will show in progress template page if completes successfully then success template otherwise failure..

>  I think we can afford to keep them around much longer than 10 minutes

We can, but the purpose of keeping the dict is to fetch the job status using the job ID. We only do that through the status page, where the status page polls the Status API to fetch the current state. Once we leave that page, we no longer have access to the job ID, so the user won’t be able to check the job status anyway. Therefore, keeping the value in dict after 10 minutes wouldn’t provide any real value to the user.

sync-page/event-handler.py
Bala-Sakabattula: if two user requests at the same the cleanup job function needs to run twice simultaneously waiting other request for longer time until the first cleanup completes ...It can cause some time delay ...

webbnh: _[Sorry for the late response, I somehow missed your reply.]_

If two user requests arrive simultaneously, then they would both call the cleanup job function; one would be forced to wait while the other did the clean-up, but, once the cleanup was done, the waiting thread would then return immediately upon finding that there was no work to do (because it had all been done by the other request).  If the other request hadn't collided with it (e.g., if it had arrived just a moment later), then the second request would have had to spend that time cleaning up anyway, so there is no net-loss in that case.

Regardless, it is true that requiring the requestor to clean up previous requests delays the response to the request.  But, I don't think the delay is significant (in fact, it's probably not even measurable in the typical case), and in exchange for imposing that delay we remove the need to create and manage a separate clean-up thread, which I think makes it absolutely worthwhile.

Bala-Sakabattula: yeah got it... I changed the cleanup job from the thread based to request based ..

PR #502, opened 2026-09-08
test addition for sync page 

Bala-Sakabattula:
Adding test case for sync page and also as per the feedback https://github.com/release-engineering/Sync2Jira/pull/501#discussion_r3927297170 calling the clean up job at requests..

### DIFF
--- a/sync-page/event-handler.py
+++ b/sync-page/event-handler.py
@@ -34,25 +34,22 @@ log = logging.getLogger("sync2jira-sync-page")
 
 
 def _cleanup_expired_jobs():
-    """Daemon thread: remove terminal jobs older than JOB_TTL_SECONDS."""
-    while True:
-        time.sleep(120)  # check every 2 minutes
-        cutoff = time.monotonic() - JOB_TTL_SECONDS
-        with _jobs_lock:
-            if not _jobs:
-                continue
-            expired = [
-                jid
-                for jid, job in _jobs.items()
-                if job["status"] in ("completed", "failed")
-                and job["finished_at"] < cutoff
-            ]
-            for jid in expired:
-                _jobs.pop(jid)
-                log.debug("Expired sync job %s from memory", jid)
-
-
-threading.Thread(target=_cleanup_expired_jobs, daemon=True, name="job-cleanup").start()
+    """Remove terminal jobs older than JOB_TTL_SECONDS."""
+    cutoff = time.monotonic() - JOB_TTL_SECONDS
+    with _jobs_lock:
+        if not _jobs:
+            log.debug("No jobs to cleanup")
+            return
+        expired = [
+            jid
+            for jid, job in _jobs.items()
+            if job["status"] in ("completed", "failed")
+            and job.get("finished_at") is not None
+            and job["finished_at"] < cutoff
+        ]
+        for jid in expired:
+            _jobs.pop(jid)
+            log.debug("Expired sync job %s from memory", jid)
 
 
 def _run_sync(job_id: str, repos: list):
@@ -85,6 +82,7 @@ def handle_event():
     Kicks off sync in a background thread and immediately returns an
     in-progress page so the gateway never times out.
     """
+    _cleanup_expired_jobs()
     response = request.form
     repos_to_sync = [repo for repo, switch in response.items() if switch == "on"]
 
--- a/tests/test_event_handler.py
+++ b/tests/test_event_handler.py
@@ -1,0 +1,289 @@
+# Build-In Modules
+import importlib.util
+import os
+import pathlib
+import sys
+import threading
+import time
+import unittest
+from unittest import mock
+
+# Set required env vars before the module is imported
+os.environ.setdefault("BASE_URL", "localhost")
+os.environ.setdefault("REDIRECT_URL", "example.com")
+
+_MODULE_PATH = str(
+    pathlib.Path(__file__).parent.parent / "sync-page" / "event-handler.py"
+)
+
+# Load the event-handler module with load_config mocked so no real config file is needed.
+with mock.patch(
+    "sync2jira.main.load_config", return_value={"sync2jira": {"map": {"github": {}}}}
+):
+    spec = importlib.util.spec_from_file_location("event_handler", _MODULE_PATH)
+    eh = importlib.util.module_from_spec(spec)
+    sys.modules["event_handler"] = eh
+    spec.loader.exec_module(eh)  # type: ignore[union-attr]
+
+PATH = "event_handler."
+
+
+def _make_job(status="in_progress", repos=None, finished_at=None, error=None):
+    """Helper to build a job dict matching the shape used by _run_sync."""
+    return {
+        "status": status,
+        "repos": repos if repos is not None else ["org/repo"],
+        "error": error,
+        "finished_at": finished_at,
+    }
+
+
+class TestHandleEvent(unittest.TestCase):
+    """Tests for the /handle-event POST endpoint."""
+
+    def setUp(self):
+        self.mock_config = {
+            "sync2jira": {"testing": False, "map": {"github": {"org/repo": {}}}}
+        }
+        eh._jobs.clear()
+        eh._jobs_repo.clear()
+        self.client = eh.app.test_client()
+
+    def test_no_repos_selected_returns_failure_page(self):
+        resp = self.client.post("/handle-event", data={})
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Failed", resp.data)
+
+    def test_all_repos_off_returns_failure_page(self):
+        resp = self.client.post("/handle-event", data={"org/repo": "off"})
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Failed", resp.data)
+
+    def test_already_syncing_same_repo_returns_failure_with_error(self):
+        with eh._jobs_repo_lock:
+            eh._jobs_repo.add("org/repo")
+        resp = self.client.post("/handle-event", data={"org/repo": "on"})
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Already syncing", resp.data)
+
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_valid_repos_returns_in_progress_page(self, _mock_pr, _mock_issues):
+        resp = self.client.post("/handle-event", data={"org/repo": "on"})
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Sync in Progress", resp.data)
+
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_valid_repos_creates_in_progress_job(self, _mock_pr, _mock_issues):
+        # Block the sync thread until we have checked the in_progress state
+        sync_started = threading.Event()
+        _mock_issues.side_effect = lambda *a, **kw: sync_started.wait()
+
+        self.client.post("/handle-event", data={"org/repo": "on"})
+
+        with eh._jobs_lock:
+            self.assertEqual(len(eh._jobs), 1)
+            job = next(iter(eh._jobs.values()))
+        self.assertEqual(job["status"], "in_progress")
+        self.assertEqual(job["repos"], ["org/repo"])
+
+        sync_started.set()  # let the thread finish cleanly
+
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_valid_repos_added_to_jobs_repo_set(self, _mock_pr, _mock_issues):
+        sync_started = threading.Event()
+        _mock_issues.side_effect = lambda *a, **kw: sync_started.wait()
+
+        self.client.post("/handle-event", data={"org/repo": "on"})
+
+        with eh._jobs_repo_lock:
+            self.assertIn("org/repo", eh._jobs_repo)
+
+        sync_started.set()
+
+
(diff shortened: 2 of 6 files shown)


sync-page/event-handler.py
```diff
@@ -34,25 +34,22 @@
...
+    with _jobs_lock:
+        if not _jobs:
+            log.debug("No jobs to cleanup")
+            return
+        expired = [
+            jid
+            for jid, job in _jobs.items()
+            if job["status"] in ("completed", "failed")
+            and job.get("finished_at") is not None
+            and job["finished_at"] < cutoff
```
webbnh, line 48: If you like, you can combine these two together by providing an explicit default value for the not-found case, e.g.:
```suggestion
            and job.get("finished_at", 1e99) < cutoff
```
That is, if there is no `"finished_at"` key, then `get()` will return a large number which should never be less than `cutoff`, resulting in the job being considered "not expired".

Note that it _is_ possible that the value returned by `time.monotonic()` could be less than `JOB_TTL_SECONDS` (e.g., during the first 10 minutes after the system initially boots), which could result in `cutoff` having a negative value.  However, this is probably not worth guarding against, as the code will still function as intended.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
+# Build-In Modules
```
webbnh, line 1: I would omit this comment line, but, if you're going to keep it, it should probably say "Built" rather than "Build".

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+)
+
+# Load the event-handler module with load_config mocked so no real config file is needed.
+with mock.patch(
+    "sync2jira.main.load_config", return_value={"sync2jira": {"map": {"github": {}}}}
+):
+    spec = importlib.util.spec_from_file_location("event_handler", _MODULE_PATH)
+    eh = importlib.util.module_from_spec(spec)
+    sys.modules["event_handler"] = eh
+    spec.loader.exec_module(eh)  # type: ignore[union-attr]
```
webbnh, line 26: Is it really necessary to perform all these steps manually?  It won't work to just import the module with the patch in place?
```suggestion
    import event_handler
```

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+_MODULE_PATH = str(
+    pathlib.Path(__file__).parent.parent / "sync-page" / "event-handler.py"
+)
+
+# Load the event-handler module with load_config mocked so no real config file is needed.
+with mock.patch(
+    "sync2jira.main.load_config", return_value={"sync2jira": {"map": {"github": {}}}}
+):
+    spec = importlib.util.spec_from_file_location("event_handler", _MODULE_PATH)
+    eh = importlib.util.module_from_spec(spec)
```
webbnh, line 24: Given that `eh` is not a _local_ variable -- it has module-scope and you take advantage of that -- it should probably have a _much_ more descriptive name, and maybe it should be in solid-caps (similar to `PATH`).

tests/test_event_handler.py
webbnh: Ugh:  is the problem here the result of the module's file name containing a hyphen??  😬  If that's the case, then we should consider renaming the module, so that these gymnastics are not required.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+            "sync2jira": {"testing": False, "map": {"github": {"org/repo": {}}}}
+        }
+        eh._jobs.clear()
+        eh._jobs_repo.clear()
+        self.client = eh.app.test_client()
+
+    def test_no_repos_selected_returns_failure_page(self):
+        resp = self.client.post("/handle-event", data={})
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Failed", resp.data)
```
webbnh, line 55: This is interesting behavior:  we "successfully" return a _failure_ page?  Should we be returning a code other than `200` in this case?  (Like `400` or `404`?)

Ditto for the all-repos-off and already-syncing cases.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def setUp(self):
+        self.mock_config = {
+            "sync2jira": {"testing": False, "map": {"github": {"org/repo": {}}}}
+        }
+        eh._jobs.clear()
+        eh._jobs_repo.clear()
+        self.client = eh.app.test_client()
+
+    def test_no_repos_selected_returns_failure_page(self):
```
webbnh, line 52: Should we be mocking `_cleanup_expired_jobs()`?

We should probably be mocking `render_template()` and  `threading.Thread()`, since they are external dependencies.  (And, if you mock `threading.Thread()`, then I think you get a mock for `thread.start()` for automagically....)

Ditto for the other scenarios....

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def test_all_repos_off_returns_failure_page(self):
+        resp = self.client.post("/handle-event", data={"org/repo": "off"})
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Failed", resp.data)
+
+    def test_already_syncing_same_repo_returns_failure_with_error(self):
+        with eh._jobs_repo_lock:
+            eh._jobs_repo.add("org/repo")
+        resp = self.client.post("/handle-event", data={"org/repo": "on"})
```
webbnh, line 65: This would be a better test if the `_jobs_repo` list only partly overlapped with the `data` list (e.g., if they each had two items but only one in common).

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn(b"Sync in Progress", resp.data)
+
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_valid_repos_creates_in_progress_job(self, _mock_pr, _mock_issues):
+        # Block the sync thread until we have checked the in_progress state
+        sync_started = threading.Event()
+        _mock_issues.side_effect = lambda *a, **kw: sync_started.wait()
```
webbnh, line 81: This mechanism seems riskier than it needs to be:  if you mock `threading.Thread()`, then you can control (or prevent) the thread creation -- then you don't need to synchronize with it or let it run or clean it up.  And, the mock gives you a point from which to observe the CUT behavior, if that's useful.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_valid_repos_added_to_jobs_repo_set(self, _mock_pr, _mock_issues):
+        sync_started = threading.Event()
+        _mock_issues.side_effect = lambda *a, **kw: sync_started.wait()
+
+        self.client.post("/handle-event", data={"org/repo": "on"})
+
+        with eh._jobs_repo_lock:
+            self.assertIn("org/repo", eh._jobs_repo)
```
webbnh, line 102: I think you can do this assertion in `test_valid_repos_creates_in_progress_job()` (it has the same setup), and dispense with this separate test case.

tests/test_event_handler.py
webbnh, line 1: It would be good if you could figure out how to make the coverage for `event-handler.py` show up in [Coveralls](https://coveralls.io/builds/81632244).  The fact that the file name has a hyphen in it may be a problem, but I assume the first issue is that it's in a different subdirectory (`sync-page`).

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+        self.assertIsNone(eh._jobs["j1"]["error"])
+
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_success_records_finished_at(self, _mock_pr, _mock_issues):
+        eh._jobs["j1"] = _make_job("in_progress")
+        eh._jobs_repo.add("org/repo")
+        before = time.monotonic()
+        eh._run_sync("j1", ["org/repo"])
+        self.assertIsNotNone(eh._jobs["j1"]["finished_at"])
```
webbnh, line 180: A better assertion would be that the value has the same type as `before`; e.g.:
```suggestion
        self.assertIs(type(eh._jobs["j1"]["finished_at"]), type(before))
```
That is, we want to be sure that we can do the comparison in the _next_ assertion, as well as the fact that it's not `None`.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+        self.assertGreaterEqual(eh._jobs["j1"]["finished_at"], before)
+
+    @mock.patch(PATH + "initialize_issues")
+    @mock.patch(PATH + "initialize_pr")
+    def test_success_releases_repos(self, _mock_pr, _mock_issues):
```
webbnh, line 185: Consider having a single "test-success" function, which does one setup and call, and then checks each assertion.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+        eh._jobs["j1"] = _make_job("in_progress")
+        eh._jobs_repo.add("org/repo")
+        eh._run_sync("j1", ["org/repo"])
+        with eh._jobs_repo_lock:
+            self.assertNotIn("org/repo", eh._jobs_repo)
```
webbnh, line 190: This would be a better test if the job consisted of multiple repos and if the `_jobs_repo` value included additional repos beyond that list (e.g., as though there were more than one sync happening simultaneously).

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+    @mock.patch(PATH + "initialize_pr")
+    def test_failure_records_finished_at(self, _mock_pr, _mock_issues):
+        eh._jobs["j1"] = _make_job("in_progress")
+        eh._jobs_repo.add("org/repo")
+        eh._run_sync("j1", ["org/repo"])
+        self.assertIsNotNone(eh._jobs["j1"]["finished_at"])
+
+    @mock.patch(PATH + "initialize_issues", side_effect=RuntimeError("err"))
+    @mock.patch(PATH + "initialize_pr")
+    def test_failure_still_releases_repos(self, _mock_pr, _mock_issues):
```
webbnh, line 211: All the failure tests have the same setup; consider combining them into a single test with multiple assertions.

Alternately, consider having one of the scenarios triggered by a failure in `initialize_pr()` instead of in `initialize_issues()`.

tests/test_event_handler.py
webbnh, line 190: Why do you need the lock for reading and not for writing?

Since this test doesn't actually create any threads (as far as I can tell), I don't think it needs a lock at all.  Ditto `test_failure_still_releases_repos()`, et seq. et al.

tests/test_event_handler.py
```diff
@@ -0,0 +1,289 @@
...
+    @mock.patch(PATH + "initialize_pr")
+    def test_multiple_repos_all_released_on_failure(self, _mock_pr, _mock_issues):
+        repos = ["org/repo-a", "org/repo-b"]
+        eh._jobs["j1"] = _make_job("in_progress", repos=repos)
+        with eh._jobs_repo_lock:
+            eh._jobs_repo.update(repos)
+        eh._run_sync("j1", repos)
+        with eh._jobs_repo_lock:
+            for r in repos:
+                self.assertNotIn(r, eh._jobs_repo)
```
webbnh, line 240: This is a valid approach, but, since `_jobs_repo` is a `set`, why not use a set operation for this assertion?  E.g.:
```suggestion
            self.assertTrue(repos.isdisjoint(eh._jobs_repo))
```

tests/test_event_handler.py
webbnh, line 1: As long as you're at it, do you want to have tests for the `"/"` and `"/github"` routes?  The tests are trivial, but they improve the coverage and help prevent "rot".  😇

webbnh (commented):
There's no single item here that I would block the merge over, but there are a number of small things (which seem to be centered around the _name_ of the module 😛) which I think you should consider before merging, and I've got a bunch of other tweaks and suggestions, as well.

tests/test_event_handler.py
```diff
@@ -12,18 +9,12 @@
 os.environ.setdefault("BASE_URL", "localhost")
 os.environ.setdefault("REDIRECT_URL", "example.com")
 
-_MODULE_PATH = str(
-    pathlib.Path(__file__).parent.parent / "sync-page" / "event-handler.py"
-)
+sys.path.insert(0, str(pathlib.Path(__file__).parent.parent / "sync-page"))
```
webbnh, line 12: This is an unsatisfying hack.  Do you understand why it is necessary?

Could we get the same effect by omitting this from here and changing the Tox or unittest configuration?  (None of the `sync2jira` tests have this...so why does the `sync-page` test need it?)

webbnh (dismissed):
This looks excellent, Bala.  There is just one item (other than my previous comment about trivial tests 😉) which I'd like you to consider.

But, if you're satisfied with both as they are, let me know and I'll close the conversations.

tests/test_event_handler.py
Bala-Sakabattula: The sync-page folder contains a hyphen (-) in its name, so it can't be imported as a normal Python package/module (unlike sync2jira, where we can do import sync2jira.downstream_issue as d).

That's why I use sys.path.insert(0, ...) to add the sync-page directory to Python's module search path. This allows me to import event_handler.py directly with:

`import event_handler as eh`

webbnh: Ugh.  OK, that makes sense, I guess.  (We really should rename the directory, too...but we can leave that for another day.)

webbnh (approved):
Looks great, @Bala-Sakabattula.

